PR #52, opened 2024-02-29
add deletion modes for indexing

akashg3627:


backend/modules/vector_db/qdrant.py
```diff
@@ -59,15 +60,102 @@ def create_collection(self, embeddings: Embeddings):
...
+                limit=self.qdrant_client.count(
+                    collection_name=self.collection_name,
+                ).count,
+                with_payload=False,
+                with_vectors=False,
+            )
+            record_ids_to_be_deleted = [record.id for record in records]
+            logger.info(f"Records to be deleted {len(record_ids_to_be_deleted)}")
+            return record_ids_to_be_deleted
+        elif deletion_mode == IndexingDeletionMode.FULL:
```
akashg3627, line 97: We should not have full mode here, it should be separate fn

PR #58, opened 2024-03-05
Added logging and comments to the code base

S1LV3RJ1NX:


backend/modules/metadata_store/models.py
```diff
@@ -8,6 +8,9 @@
 
 
 class CollectionIndexerJobRunStatus(str, enum.Enum):
+    """
```
akashg3627, line 11: we can delete this file

akashg3627 (approved):

PR #65, opened 2024-03-08
Fixed bugs to enable collection creation, embedding and data ingestion

S1LV3RJ1NX:
Fixed bugs: 
- indexer.py -> logger.info docs_to_index_count is int, removed len() 
- mlfoundryloader.py -> removed os.path.join(dest_dir, files)
- mlfoundry.py -> logger.info to get len(data_ingestion_runs) which is a generator, converted it to list then calculated the length
- vector_db/qdrant.py -> imported Qdrant from langchain.vectorstores rather than langchain_commmunity

### DIFF
--- a/backend/modules/metadata_store/mlfoundry.py
+++ b/backend/modules/metadata_store/mlfoundry.py
@@ -434,12 +434,12 @@ class MLFoundry(BaseMetadataStore):
                 filter_string=f"params.entity_type = '{MLRunTypes.DATA_INGESTION_RUN.value}' and params.collection_name = '{collection_name}'",
             )
             logger.debug(
-                f"[Metadata Store] Found {len(data_ingestion_runs)} data ingestion runs for {collection_name} to delete"
+                f"[Metadata Store] Found {len(list(data_ingestion_runs))} data ingestion runs for {collection_name} to delete"
             )
             for collection_inderer_job_run in data_ingestion_runs:
                 collection_inderer_job_run.delete()
             logger.debug(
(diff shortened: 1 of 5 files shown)


backend/modules/metadata_store/mlfoundry.py
```diff
@@ -434,12 +434,12 @@ def delete_collection(self, collection_name: str, include_runs=False):
                 filter_string=f"params.entity_type = '{MLRunTypes.DATA_INGESTION_RUN.value}' and params.collection_name = '{collection_name}'",
             )
             logger.debug(
-                f"[Metadata Store] Found {len(data_ingestion_runs)} data ingestion runs for {collection_name} to delete"
+                f"[Metadata Store] Found {len(list(data_ingestion_runs))} data ingestion runs for {collection_name} to delete"
```
akashg3627, line 437: this would cause runs to get empty as its iterator

backend/modules/query_controllers/default/controller.py
```diff
@@ -82,7 +82,10 @@ async def answer(self, request: DefaultQueryInput):
 
             # Get the answer
             logger.info(f"Request query: {request.query}")
-            outputs = qa({"query": request.query})
+
+            # Can't use ainvoke here as Qdrant uses grpc for the retrieval, 
+            # which terminates as the client grpc is set to false
```
akashg3627, line 87: not sure, but for me this worked

S1LV3RJ1NX: It is  due to using Qdrant from langchain as opposed to from langchain_community.

PR #72, opened 2024-03-15
Added support for OPENAI_API_KEY

S1LV3RJ1NX:


backend/modules/query_controllers/default/controller.py
```diff
@@ -37,14 +38,14 @@ async def answer(self, request: DefaultQueryInput):
 
             vector_store = VECTOR_STORE_CLIENT.get_vector_store(
                 collection_name=collection.name,
-                embeddings=get_embedder(collection.embedder_config),
+                embeddings=OpenAIEmbeddings(
```
akashg3627, line 41: use get_embedder function only and instead of truefoundry embedding use this class

backend/settings.py
```diff
@@ -61,6 +63,10 @@ class Settings(BaseSettings):
         )
     except Exception as e:
         raise ValueError(f"METADATA_STORE_CONFIG is invalid: {e}")
+    
+    if OPENAI_API_KEY == "" and TFY_API_KEY != "" and TFY_LLM_GATEWAY_URL != "":
```
akashg3627, line 67: check line 48-54, there are checks for api key to throw error

akashg3627 (approved):

PR #88, opened 2024-03-21
Added support for opensource LLMs, embeddings and reranking

S1LV3RJ1NX:
- Added Ollama Support.
- Adding SOTA open source embeddings from `mixedbread-ai` that are comparable to `openai-embeddings-v3`
- Added a new `document reranking` module that fetches top k documents from the retrieved ones to form a meaningful context.
- Query controller now has an examples folders under which two types of controllers are written one using openai and one fully open source
- Similary, for `local` folder
- Updated the ReadMe correspondingly.


.env.example
```diff
@@ -5,18 +5,22 @@ VECTOR_DB_CONFIG='{"provider":"qdrant","local":true}'
 # Using hosted qdrant and mlfoundry
 METADATA_STORE_CONFIG='{"provider":"mlfoundry","config":{"ml_repo_name":"<your_ml_repo_name>"}}'
 VECTOR_DB_CONFIG='{"provider":"qdrant","url":"<your_qdrant_url>"}'
+
+# Should be set to the root path of the service
+TFY_SERVICE_ROOT_PATH='/'
+
+
+# Optional as per requirement
+TFY_API_KEY="<your_truefoundry_api_key>"
```
innoavator, line 14: Let's add a line for each of these environment variables regarding what they do

local/query_opensrc.py
```diff
@@ -0,0 +1,99 @@
+
```
innoavator, line 1: We shouldn't have a separate query controller for opensource - it should be an argument in the answer function

backend/modules/reranker/mxbai_reranker.py
```diff
@@ -0,0 +1,31 @@
+from langchain.retrievers.document_compressors.base import BaseDocumentCompressor
+from langchain.docstore.document import Document
+from typing import List, Optional, Sequence
+from langchain.callbacks.manager import Callbacks
+from sentence_transformers import CrossEncoder
+
```
innoavator, line 6: Add the blog link here about the MxBaiReranker and why it should be used.

backend/server/services/collection.py
```diff
@@ -43,6 +45,8 @@ def create_collection(collection: CreateCollectionDto):
                     embedder_config=collection.embedder_config,
                 )
             )
+
+            print(created_collection.dict())
```
innoavator, line 49: remove the print line

innoavator (approved):
Approved - but please address the changes before merging

PR #96, opened 2024-03-27
feat: add retrievers, model configs, parser maps

sayan-truefoundry:
image

image

image

### DIFF
--- a/frontend/src/screens/dashboard/docsqa/AddDataSourceToCollection.tsx
+++ b/frontend/src/screens/dashboard/docsqa/AddDataSourceToCollection.tsx
@@ -2,6 +2,7 @@ import Button from '@/components/base/atoms/Button'
 import CustomDrawer from '@/components/base/atoms/CustomDrawer'
 import Spinner from '@/components/base/atoms/Spinner/Spinner'
 import notify from '@/components/base/molecules/Notify'
+import SimpleCodeEditor from '@/components/base/molecules/SimpleCodeEditor'
 import {
   Collection,
   useAddDocsToCollectionMutation,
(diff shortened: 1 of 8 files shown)


frontend/src/screens/dashboard/docsqa/AddDataSourceToCollection.tsx
```diff
@@ -12,9 +13,13 @@ import { MenuItem, Select } from '@mui/material'
 import React, { useState } from 'react'
 
 const defaultParserMaps = `{
```
sajal-truefoundry, line 15: name of variable should be different, it's not just for parser map now

frontend/src/screens/dashboard/docsqa/index.tsx
```diff
@@ -204,8 +179,8 @@ const DocsQA = () => {
                   sx={{
                     background: 'white',
                     height: '32px',
-                    width: '13rem',
-                    minWidth: '13rem',
+                    width: '13.2rem',
```
sajal-truefoundry, line 182: Try to use rem values that are equivalent to a whole pixel.

Considering `16px` to be the root size, rem size should be a multiple of `0.0625`

frontend/src/screens/dashboard/docsqa/index.tsx
```diff
@@ -274,11 +249,74 @@ const DocsQA = () => {
                 onChange={(e) => setPromptTemplate(e.target.value)}
               />
             </div>
+            <div className="h-full border rounded-lg border-[#CEE0F8] w-[calc(100%-400px)] bg-white p-4">
+              <div className="flex gap-4 items-center">
+                <div className="w-full relative">
+                  <Input
+                    className="w-[100%] h-[44px] text-sm pr-14"
```
sajal-truefoundry, line 256: - `w-[100%]` is not needed as `w-full` is already defined in tailwind
- Please don't use pixel values, only use rems

sajal-truefoundry (changes requested):

PR #110, opened 2024-04-03
Refactoring of dataloaders, embedders and requirements update

S1LV3RJ1NX:
1. Removed services and added the functionality to routers
2. Updated ingest.py to use router ingest
3. Clubed together tf artifact and data-dir as truefoundry dataloader
4. Renamed local dataloader localdir dataloader and added description to all dataloaders
5. Updated backend/requirements.txt with truefoundry[ml]==0.1.1
6. Modified all imports that use servicefoundry to use truefoundry
7. Added LOCAL env var (future use) for using local models, etc
8. Embedder provider default changed to openai and truefoundry consistent with dataloaders provider


backend/modules/dataloaders/truefoundryloader.py
```diff
@@ -0,0 +1,115 @@
...
+            logger.info("Downloading Artifact: {}".format(artifact_fqn))
+
+            # Get the artifact version directly
+            dataset = mlfoundry_client.get_artifact_version_by_fqn(artifact_fqn)
+            # download it to disk
+            # `download_info` points to a directory that has all contents of the artifact
+            download_info = dataset.download(path=dest_dir)
+            logger.debug(f"Artifact download info: {download_info}")
+
+        elif data_source.uri.startswith("data_directory"):
```
akashg3627, line 46: i think its `data_dir`

local/ingest.py
```diff
@@ -6,11 +6,8 @@
 
 # Data ingestion
 from backend.modules.metadata_store.client import METADATA_STORE_CLIENT
-from backend.modules.vector_db.client import VECTOR_STORE_CLIENT
-from backend.types import IngestDataToCollectionDto, DataIngestionMode
-from backend.server.services.collection import CollectionService
-from backend.modules.embedder.embedder import get_embedder
-
+from backend.types import IngestDataToCollectionDto
+from backend.server.routers.collection import ingest_data
```
akashg3627, line 10: we need to refactor this. Lets shift the function to indexer folder and let backend router import it instead

akashg3627: check indexer.main.py

backend/modules/dataloaders/truefoundryloader.py
S1LV3RJ1NX: fixed.

local/ingest.py
S1LV3RJ1NX: fixed.

akashg3627 (approved):

PR #111, opened 2024-04-03
Updated models API to list local models and embeddings

S1LV3RJ1NX:


backend/server/routers/internal.py
```diff
@@ -45,6 +46,56 @@ async def upload_to_data_directory(req: UploadToDataDirectoryDto):
...
+                url = "https://api.openai.com/v1/models"
+                headers = {"Authorization": f"Bearer {settings.OPENAI_API_KEY}"}
+                response = requests.get(url=url)
+                data = response.json()
+                # TODO: Add the models to the enabled_models list
+            except Exception as ex:
+                logger.error(f"Error fetching openai models: {ex}")
+
+    # Models from the llm gateway
     url = f"{settings.TFY_HOST}/api/svc/v1/llm-gateway/model/enabled"
```
akashg3627, line 99: we should add a check for `TFY_API_KEY` before fetching these models

akashg3627 (approved):

backend/server/routers/internal.py
```diff
@@ -45,21 +46,76 @@ async def upload_to_data_directory(req: UploadToDataDirectoryDto):
...
+            response.raise_for_status()
+        
+            data: dict[str, dict[str, list[dict]]] = response.json()
+
+            for provider_accounts in data.values():
+                for models in provider_accounts.values():
+                    for model in models:
+                        if model_type is None or model_type in model["types"]:
+                            # for models from llm gateway backend_provider is truefoundry
+                            model["backend_provider"] = "truefoundry"
```
akashg3627, line 113: where are we using this field?

PR #132, opened 2024-04-11
Added support for cloning qdrant db across different/same qdrant instance

S1LV3RJ1NX:


backend/migration/qdrant_migration.py
```diff
@@ -0,0 +1,163 @@
...
+        ).dict()
+
+        
+        logger.debug(f"Creating '{dest_collection.get('name')}' collection at destination")
+        # create collection at destination
+        with requests.post(
+                url = f"{destination_backend_url.rstrip('/')}/v1/collections/",
+                json = dest_collection
+            ) as r:
+            logger.debug("Collection enrty created: ", r.json())
```
chiragjn, line 80:
```suggestion
            logger.debug("Collection entry created: ", r.json())
```

backend/migration/qdrant_migration.py
```diff
@@ -0,0 +1,163 @@
...
+        raise Exception("No collections found")
+
+    fetched_collection = None
+    for collection in collections:
+        if collection.get("name") == source_collection_name:
+            fetched_collection = collection
+            break
+
+    if fetched_collection == None:
+        raise Exception(f"Collection {source_collection_name} not found")
```
chiragjn, line 54: Why not 
https://qdrant.github.io/qdrant/redoc/index.html#tag/collections/operation/get_collection 
?

backend/migration/qdrant_migration.py
```diff
@@ -0,0 +1,163 @@
...
+    source_prefix: str,
+    destination_backend_url: str,
+    destination_collection_name: str,  
+    destination_qdrant_url: str,
+    destination_prefix: str, 
+):
+    # fetch collection from source
+    collections = None
+    with requests.get(f"{source_backend_url.rstrip('/')}/v1/collections/") as r:
+        collections = r.json().get("collections")
```
chiragjn, line 42: Weird, we don't have get on a single collection yet 😅 
Let's add it to backlog

backend/migration/qdrant_migration.py
```diff
@@ -0,0 +1,163 @@
...
+
+    if collections == None:
```
chiragjn, line 44:
```suggestion
    if not collections:
```

backend/migration/qdrant_migration.py
```diff
@@ -0,0 +1,163 @@
...
+        raise Exception("No collections found")
+
+    fetched_collection = None
+    for collection in collections:
+        if collection.get("name") == source_collection_name:
+            fetched_collection = collection
+            break
+
+    if fetched_collection == None:
```
chiragjn, line 53:
```suggestion
    if fetched_collection is None:
```

backend/migration/utils.py
```diff
@@ -0,0 +1,131 @@
...
+        batch_size (int, optional): Batch size for scrolling and uploading vectors. Defaults to 100.
+    """
+    if _has_custom_shards(source_client, source_collection_name):
+        raise ValueError("Migration of collections with custom shards is not supported yet")
+    
+    if source_collection_name == destination_collection_name and same_qdrant:
+        destination_collection_name = destination_collection_name + "copy"
+        logger.debug(f"Destination collection name is same as source collection name. Renaming destination collection to {destination_collection_name}")
+
+    _recreate_collection(
```
chiragjn, line 37: What if destination qdrant is different but destination collection exists? we should either break or just accept the collection as it instead of trying to re-create it

backend/migration/utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Args:
+        source_collection_name (str): Collection name of source
+        destination_collection_name (str): Collection name of destination
+        source_client (QdrantBase): Source client
+        dest_client (QdrantBase): Destination client
+        batch_size (int, optional): Batch size for scrolling and uploading vectors. Defaults to 100.
+    """
+    records, next_offset = source_client.scroll(source_collection_name, limit=2, with_vectors=True)
+    dest_client.upload_points(destination_collection_name, records, wait=True)  # type: ignore
```
chiragjn, line 117: odd, why is this needed? specifically `limit=2` ?

backend/migration/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    while next_offset is not None:
+        records, next_offset = source_client.scroll(
+            source_collection_name, offset=next_offset, limit=batch_size, with_vectors=True
+        )
+        dest_client.upload_points(destination_collection_name, records, wait=True)
+    source_client_vectors_count = source_client.get_collection(source_collection_name).vectors_count
+    dest_client_vectors_count = dest_client.get_collection(destination_collection_name).vectors_count
+    assert (
+        source_client_vectors_count == dest_client_vectors_count
+    ), f"Migration failed, vectors count are not equal: source vector count {source_client_vectors_count}, dest vector count {dest_client_vectors_count}"
```
chiragjn, line 131: This should probably only be a warning

backend/migration/qdrant_migration.py
```diff
@@ -0,0 +1,163 @@
...
+    source_backend_url = "http://localhost:8000"
+    source_collection_name = "creditcard"
+    source_qdrant_url = "https://ps-qdrant-prathamesh-ws.devtest.truefoundry.tech"
+    source_prefix = None
+
+    # DESTINATION
+    destination_backend_url = "http://localhost:8000"
+    destination_collection_name = "cloned"
+    destination_qdrant_url = "https://ps-qdrant-prathamesh-ws.devtest.truefoundry.tech"
+    destination_prefix = None
```
chiragjn, line 161: minor: cleanup

chiragjn (approved):
Generally looks good, added a few suggestions

backend/migration/utils.py
S1LV3RJ1NX: ok, we arent checking if the destination qdrant exists or not yet. we might need to do that before migration. How about setting an override flag? If that flag is set, and destination collection exists, we recreate else we break?

backend/migration/utils.py
S1LV3RJ1NX: This is just to get the initial offset. It can be any number I have just kept it as a small number. To iterate over other points later if any.

backend/migration/utils.py
S1LV3RJ1NX: done

PR #177, opened 2024-05-29
Fixed data source visiblity on FE

S1LV3RJ1NX:


backend/modules/metadata_store/local.py
```diff
@@ -106,8 +106,14 @@ def get_data_source_from_fqn(self, fqn: str = None) -> DataSource | None:
     def get_data_sources(self) -> List[DataSource]:
         return [self.data_source]
 
-    def list_data_sources(self) -> List[str]:
-        return [self.data_source.fqn]
+    async def list_data_sources(self) -> List[str]:
+        return [
+            {
+                "type": self.data_source.fqn.split("::")[0],
+                "uri": self.data_source.fqn.split("::")[1],
```
chiragjn, line 113: Why do we need to split like this?
Do we not have type and uri separately on the data_source instance itself?

S1LV3RJ1NX: Right!, Thanks for bringing it to notice

Issue #191, opened 2024-05-30
Feature Request: Publish Cognita as a package to pypi


avibathula:
For higher adoption, you have to have stable releases of Cognita and publish it as a package to `pypi`

This is another aspect [SchiPhi's R2R](https://github.com/SciPhi-AI/R2R/tree/main) is taking a lead w.r.t meeting developer expectations.

And please publish a roadmap with prioritized items.

Issue #190, opened 2024-05-30
Switch from `pip` based dependency management to `poetry`


avibathula:
https://realpython.com/dependency-management-python-poetry/

chiragjn:
Thanks for brining this up,
We have it on our plans to do start using poetry and publish cognita
That will also make installing extra optional packages straighforward and allow use to depend on ranges of packages

We were mostly waiting to reach a stable state on our interfaces before we start being a package with proper semantic versioning

PR #194, opened 2024-05-30
Use poetry for dependencies management

chiragjn:
See: #190

Starting this is an intial draft to move to poetry

TODO:
- [ ] Test Dockerfile changes
- [ ] Make cognita a publishable package (#191) - We do not want our users to worry about poetry
    - [ ] Refactor to put source code in a `cognita` module - Changing imports
    - [ ] Add cli to invoke serve and index - Update entrypoints in pyproject.toml
    - [ ] Update Dockerfile
    - [ ] Remove requirements.txt
- [ ] Add a workflow to publish package
- [ ] Update Docs

Later:
- [ ] Add a workflow to publish docker images
- [ ] Setup pytest and add a workflow


backend/pyproject.toml
```diff
@@ -0,0 +1,75 @@
+[tool.poetry]
+name = "cognita"
+version = "0.0.0"
+description = "Cognita"
+authors = ["Abhishek Choudhary <abhichoudhary06@gmail.com>"]
+readme = "README.md"
```
chiragjn, line 6: This does not exist relative to this file
We want to move pyproject.toml to root

chiragjn:
I'll re-open this later, we already have conflicts with recent changes

PR #195, opened 2024-05-30
Docker compose and fixes

S1LV3RJ1NX:


docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
+version: "3.8"
+services:
+    ollama-server:
+        image: ollama/ollama:0.1.39
+        pull_policy: always
+        restart: always
+        container_name: ollama
+        ports:
+            - "${OLLAMA_PORT}:11434"
```
chiragjn, line 9: exposing ports to host is not strictly necessary if we are using a bridge network
Should only need to expose for backend and frontend

docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+            - .:/app
+        ports:
+            - "${COGNITA_FRONTEND_PORT}:5000"
+        depends_on:
+            - ollama-server
+            - infinity-server
+            - qdrant-server
+            - cognita-backend
+        environment:
+            - VITE_QA_FOUNDRY_URL=http://host.docker.internal:${COGNITA_FRONTEND_PORT}
```
chiragjn, line 122: Is this supposed to be backend url or frontend url?

docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
+version: "3.8"
+services:
+    ollama-server:
+        image: ollama/ollama:0.1.39
+        pull_policy: always
```
chiragjn, line 5: why always?

docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+        restart: always
```
chiragjn, line 6: unless-stopped is a better alternative imo

docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+            - "${INFINITY_PORT}:7997"
+        volumes:
+            - ./volumes/infinity:/app/.cache
+        healthcheck:
+            test:
+                [
+                    "CMD",
+                    "curl",
+                    "-f",
+                    "http://localhost:${INFINITY_PORT}/health",
```
chiragjn, line 40: Afaik healthchecks are executed inside the container so this should be port inside the container on which infinity is listening to
same for other healthchecks

docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+        networks:
+            - cognita-docker
+
+    infinity-server:
+        image: michaelf34/infinity:0.0.36
+        pull_policy: always
+        restart: always
+        container_name: infinity
+        ports:
+            - "${INFINITY_PORT}:7997"
```
chiragjn, line 31: does infinity run on 7997 by default?

docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+    qdrant-server:
+        image: qdrant/qdrant:v1.8.4
+        pull_policy: always
+        restart: always
+        container_name: qdrant
+        ports:
+            - "${QDRANT_REST}:6333"
+            - "${QDRANT_GRPC}:6334"
+        volumes:
+            - ./volumes/qdrant_storage:/qdrant/storage:z
```
chiragjn, line 58: why does this need `z` mode?

.dockerignore
```diff
@@ -20,3 +20,6 @@ tf-docs-crawled/
 truefoundry-docs/
 *.zip
 ragenv/
+volumes/
+qdrant_db/
```
chiragjn, line 24: Arent qdrant ones part of volumes/ folder which is already ignore

compose.env
```diff
@@ -0,0 +1,38 @@
...
+INFINITY_PORT=7997
+
+## QDRANT
+QDRANT_REST=6333
+QDRANT_GRPC=6334
+
+## COGNITA_BACKEND VARS
+COGNITA_BACKEND_PORT=8000
+COGNITA_BACKEND_HOST=0.0.0.0
+OLLAMA_URL=http://host.docker.internal:${OLLAMA_PORT}
```
chiragjn, line 21: for all _HOST / _URL we can use the bridge network instead of relying on host.docker.internal

docker-compose.yaml
S1LV3RJ1NX: yes 7999 is default

.dockerignore
S1LV3RJ1NX: This one was when we start a separate qdrant container. Just added that to ignore.

docker-compose.yaml
S1LV3RJ1NX: this should be backend. thanks for pointing out.

docker-compose.yaml
S1LV3RJ1NX: `if-not-present` works? I guess the version will not get updated anyways. will change this.

docker-compose.yaml
S1LV3RJ1NX: This was in qdrant docs, this is what I found:

:z: The SELinux label is set so that all containers can read/write the same volumes. This is typically used when multiple containers need to share a volume.

docker-compose.yaml
S1LV3RJ1NX: done.

PR #206, opened 2024-06-05
Fixed qdrant migration on update of qdrant-client

S1LV3RJ1NX:


backend/migration/utils.py
```diff
@@ -154,20 +158,19 @@ def _migrate_collection(
...
         )
         dest_client.upload_points(destination_collection_name, records, wait=True)
-    source_client_vectors_count = source_client.get_collection(
-        source_collection_name
-    ).vectors_count
-    dest_client_vectors_count = dest_client.get_collection(
-        destination_collection_name
-    ).vectors_count
+    source_client_vectors_count = source_client.count(source_collection_name).count
+    dest_client_vectors_count = dest_client.count(destination_collection_name).count
```
chiragjn, line 173: I think we should be using `dest_client.get_collection(...).points_count`

backend/migration/utils.py
```diff
@@ -94,7 +95,9 @@ def _recreate_collection(
     src_config = src_collection_info.config
     src_payload_schema = src_collection_info.payload_schema
 
-    dest_client.recreate_collection(
+    # delete destination collection only from qdrant that was created while creating metadatastore entry
+    dest_client.delete_collection(destination_collection_name, timeout=300)
```
chiragjn, line 99: Is there a `wait` option?

backend/migration/qdrant_migration.py
chiragjn, line 1: Update qdrant-client in this PR itself

backend/migration/utils.py
S1LV3RJ1NX: https://github.com/qdrant/qdrant-client/blob/d18cb1702f4cf8155766c7b32d1e4a68af11cd6a/qdrant_client/qdrant_remote.py#L1272

This returns`count`.

backend/migration/qdrant_migration.py
S1LV3RJ1NX: Actually I merged the dependabot into main missed that.

backend/migration/utils.py
S1LV3RJ1NX: don't see any wait. might add sleep?

backend/migration/qdrant_migration.py
```diff
@@ -217,7 +223,7 @@ def main():
         destination_qdrant_url=args.destination_qdrant_url,
         destination_prefix=args.destination_prefix,
         batch_size=args.batch_size,
-        overwrite=args.overwrite,
+        overwrite=args.overwrite == "True",
```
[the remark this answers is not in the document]
S1LV3RJ1NX: Will this be fine?
```
parser.add_argument(
        "--overwrite",
        type=str,
        help="Overwrite destination collection if exists in separate qdrant",
        required=False,
        default="True",
    )
```

backend/migration/utils.py
chiragjn: I was talking about `.get_collection` and not `.count`

https://github.com/qdrant/qdrant-client/blob/d18cb1702f4cf8155766c7b32d1e4a68af11cd6a/qdrant_client/http/models/models.py#L169

Anyway both should be equivalent, so it is okay

S1LV3RJ1NX: yaa..

backend/migration/utils.py
chiragjn: I supppose timeout=None is equivalent to waiting till completion

backend/migration/qdrant_migration.py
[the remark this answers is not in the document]
S1LV3RJ1NX: Right!
```
parser.add_argument(
        "--overwrite",
        type=bool,
        help="Overwrite destination collection if exists in separate qdrant",
        required=False,
        action="store_true",
    )

    args = parser.parse_args()

    migrate_collection(
        # source args
        source_backend_url=args.source_backend_url,
        source_collection_name=args.source_collection_name,
        source_qdrant_url=args.source_qdrant_url,
        source_prefix=args.source_prefix,
        # destination args
        destination_backend_url=args.destination_backend_url,
        destination_collection_name=args.destination_collection_name,
        destination_qdrant_url=args.destination_qdrant_url,
        destination_prefix=args.destination_prefix,
        batch_size=args.batch_size,
        overwrite=args.overwrite,
    )
```

Via this, if I specify `--overwrite` then it will result as `True` else `False`, right?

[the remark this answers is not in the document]
chiragjn: correct

backend/migration/utils.py
S1LV3RJ1NX: The default is set to 5sec for REST.  So, 300 should do.
```
timeout:
            Timeout for REST and gRPC API requests.
            Default: 5 seconds for REST and unlimited for gRPC
```

chiragjn (approved):

PR #212, opened 2024-06-10
Made API routes consistent in terms of trailing slashes

S1LV3RJ1NX:


backend/server/routers/collection.py
```diff
@@ -19,7 +19,7 @@
 router = APIRouter(prefix="/v1/collections", tags=["collections"])
 
 
-@router.get("/")
+@router.get("")
```
chiragjn, line 22: We might need to have both in the interrim till frontend moves

```
@router.get("")
@router.get("/")
```

chiragjn (approved):

PR #222, opened 2024-06-14
Added models_config and gateway to make it easy to switch betwe…

innoavator:
…en models


backend/modules/model_gateway/model_gateway.py
```diff
@@ -0,0 +1,80 @@
...
+        for key in self.modelsToProviderMap:
+            print(key)
+        if model_config.name not in self.modelsToProviderMap:
+            raise ValueError(f"Model {model_config.name} not registered in the model gateway.")
+        model_provider_config: ModelProviderConfig = self.modelsToProviderMap[model_config.name]
+        if not model_config.parameters:
+           model_config.parameters = {} 
+        return ChatOpenAI(
+                model=model_config.name.split("/")[1],
+                temperature=model_config.parameters.get("temperature", 0.1),
```
S1LV3RJ1NX, line 69: Maybe add an optional `model_kwargs`

backend/modules/query_controllers/example/payload.py
```diff
@@ -27,7 +26,6 @@
     "query": "Explain in detail different categories of credit cards",
     "model_configuration": {
         "name": "openai-main/gpt-3-5-turbo",
-        "provider": "truefoundry",
         "parameters": {"temperature": 0.1},
```
S1LV3RJ1NX, line 29: What do think on changing this payload similar to ChatOpenAI?
```
model_config: {
- model_name
- provider
- temperature
- max_tokens
- model_kwargs
}
```

temperature and max_tokens can go within parameters if required?

backend/types.py
```diff
@@ -90,14 +90,34 @@ class LoadedDataPoint(DataPoint):
...
+    name: str
+    type: Optional[ModelType]
+    parameters: Optional[Dict[str, Any]] = None
+    
+class ModelProviderConfig(BaseModel):
+    provider_name: str
+    api_format: str
+    llm_model_ids: List[str]
+    embedding_model_ids: List[str]
+    api_key_env_var: str
```
S1LV3RJ1NX, line 112: how is this `api_key` set?

models_config.json
```diff
@@ -0,0 +1,25 @@
+[
+  {
+    "provider_name": "openai",
+    "api_format": "openai",
+    "llm_model_ids": ["gpt-4", "gpt-3-5-turbo"],
+    "embedding_model_ids": ["text-embedding-3-small"],
+    "api_key_env_var": "OPENAI_API_KEY"
```
S1LV3RJ1NX, line 7: Not able to find, where do we set API keys? Either in the Gateway Class, initialize these vars from settings.py?

backend/settings.py
```diff
@@ -23,6 +25,8 @@ class Settings(BaseSettings):
     EMBEDDING_CACHE_CONFIG: Optional[EmbeddingCacheConfig] = None
 
     LOG_LEVEL = os.getenv("LOG_LEVEL", "info")
+    AI_GATEWAY = os.getenv("AI_GATEWAY", "openai")
+    AI_GATEWAY_CONFIG = json.loads(os.getenv("AI_GATEWAY_CONFIG", "{}"))
```
S1LV3RJ1NX, line 29: Where is this used?

backend/types.py
innoavator: You can set the env variable for the key mentioned as the value here.

models_config.json
innoavator: Its set in the gateway class - check the get_embedder and get_llm function

S1LV3RJ1NX (approved):

PR #224, opened 2024-06-17
Port pydantic v1 models to pydantic v2 

### LINKED ISSUE #219: Port pydantic v1 models to pydantic v2

abhinavk454:
Closes #219 

TODO:
- [x] Change fields which start with `model_*`, especially `model_config` - this may or may not be a breaking change
- [x] Verify if prisma store is still able to function correctly


backend/settings.py
```diff
@@ -26,14 +26,14 @@ class Settings(BaseSettings):
     VECTOR_DB_CONFIG = os.getenv("VECTOR_DB_CONFIG", "")
     METADATA_STORE_CONFIG = os.getenv("METADATA_STORE_CONFIG", "")
     TFY_SERVICE_ROOT_PATH = os.getenv("TFY_SERVICE_ROOT_PATH", "")
-    JOB_FQN = os.getenv("JOB_FQN", "")
-    JOB_COMPONENT_NAME = os.getenv("JOB_COMPONENT_NAME", "")
+    JOB_FQN:ClassVar[str] = os.getenv("JOB_FQN", "")
+    JOB_COMPONENT_NAME:ClassVar[str] = os.getenv("JOB_COMPONENT_NAME", "")
```
chiragjn, line 30: technically all of these can be class var, but why are we specifically changing these?

backend/settings.py
```diff
@@ -1,10 +1,10 @@
...
+from typing import Optional,ClassVar
 
 import orjson
-from pydantic import BaseSettings
 
 from backend.types import EmbeddingCacheConfig, MetadataStoreConfig, VectorDBConfig
+from pydantic_settings import BaseSettings
 
 
 class Settings(BaseSettings):
```
chiragjn, line 10: Maybe not in this PR but I think we should be able to get rid of `os.getenv`
after all pydantic settings is supposed to read from env on its own

backend/server/decorators.py
```diff
@@ -59,7 +58,7 @@ def _init_cbv(cls: Type[Any]) -> None:
     ]
     dependency_names: List[str] = []
     for name, hint in get_type_hints(cls).items():
-        if is_classvar(hint):
+        if getattr(hint, "__origin__", None) is ClassVar:
```
chiragjn, line 61: I'll have to look this up in the docs somewhere but probably this is correct

backend/settings.py
abhinavk454: These 2 were not present in env so need to annotate these, if not done they were throwing error:

PydanticUserError(pydantic.errors.PydanticUserError: A non-annotated attribute was detected: JOB_FQN = ''. All model fields require a type annotation; if JOB_FQN is not meant to be a field, you may be able to resolve this error by annotating it as a ClassVar or updating model_config['ignored_types'].

chiragjn:
Apologies for the delay here, we had some improvements planned which took higher priority.
I'll pick this up now and resolve the conflicts

backend/database/schema.prisma
```diff
@@ -24,6 +24,7 @@ model Collection {
     description             String?
     embedder_config         Json
     // Collection can have multiple data sources
+    // TODO (chiragjn): Why does this have to be Nullable? Default to {}
```
S1LV3RJ1NX, line 27: While creating collection, user might not provide a datasource via API. Hence initially it can be None

S1LV3RJ1NX (commented):
Answered few of your answer.

backend/modules/metadata_store/base.py
```diff
@@ -365,17 +368,19 @@ async def alist_collections(
             self.list_collections,
         )
 
+    # TODO (chiragjn): Return complete entities, why return dict?
     def list_data_sources(
         self,
-    ) -> List[str]:
+    ) -> List[Dict[str, str]]:
```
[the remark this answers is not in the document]
chiragjn: The implementation was returning dicts, type hint was wrong, I have not changed the functionality

backend/modules/metadata_store/base.py
```diff
@@ -354,6 +356,7 @@ def list_collections(
         """
         raise NotImplementedError()
 
+    # TODO (chiragjn): Return complete entities, why return only str?
```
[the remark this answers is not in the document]
S1LV3RJ1NX: Agree, but this was done as mlfoundry took a lot of time to return entire entities. You can check the difference in the list_collections and get_collections API of truefoundry mdstore.

chiragjn (approved):

backend/database/schema.prisma
S1LV3RJ1NX: The reason being here, I checked, first the collection is created whose pydantic `CreateCollection` does not have associated data sources, then the source is added later in the API.

backend/types.py
```diff
@@ -331,7 +324,14 @@ class CreateDataSource(BaseDataSource):
...
 class DataSource(BaseDataSource):
-    pass
+    @model_serializer
+    def serialize(self):
+        return {
+            "type": self.type,
+            "uri": self.uri,
+            "metadata": self.metadata,
+            "fqn": self.fqn,
+        }
```
chiragjn, line 334: We do not need this

backend/types.py
```diff
@@ -139,37 +146,23 @@ class EmbedderConfig(ModelConfig):
...
-    )
+    type: ModelType = ModelType.parser
+
+    @model_serializer
+    def serialize(self):
+        return {
+            "name": self.name,
+            "parameters": self.parameters,
+            "type": self.type.value,
+        }
```
chiragjn, line 165: should not need this

backend/types.py
```diff
@@ -139,37 +146,23 @@ class EmbedderConfig(ModelConfig):
...
-    chunk_overlap: int = Field(title="Chunk Overlap for indexing", ge=0, default=20)
-    parser_map: Dict[str, str] = Field(
-        title="Mapping of file extensions to parsers",
-        default_factory=dict,
-    )
-    additional_config: Optional[Dict[str, Any]] = Field(
-        title="Additional optional configuration for the parser",
-        default_factory=dict,
-    )
+    type: ModelType = ModelType.parser
```
chiragjn, line 157: We probably need use_enum_values equivalent behavior to serialize correctly, check once
We need to check all models where we use enums

backend/types.py
```diff
@@ -120,7 +120,14 @@ class ModelConfig(BaseModel):
...
     type: Optional[ModelType] = None
-    parameters: Dict[str, Any] = Field(default_factory=dict)
+    parameters: Optional[Dict[str, Any]] = Field(default_factory=dict)
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "name": self.name,
+            "type": self.type,
+            "parameters": self.parameters,
+        }
```
chiragjn, line 130: Is this really needed?
model_dump() should be enough

backend/modules/metadata_store/prismastore.py
```diff
@@ -255,18 +257,19 @@ async def aassociate_data_source_with_collection(
...
                 }
 
+            logger.info(existing_collection_associated_data_sources)
             associated_data_sources: Dict[str, Dict[str, Any]] = {}
             for (
                 data_source_fqn,
                 data_source,
             ) in existing_collection_associated_data_sources.items():
-                associated_data_sources[data_source_fqn] = data_source.model_dump()
+                associated_data_sources[data_source_fqn] = data_source.serialize()
```
chiragjn, line 272: Why serialize() and not model_dump() ?

Also what if some item is already a dict?

backend/modules/metadata_store/prismastore.py
```diff
@@ -337,6 +340,11 @@ async def aunassociate_data_source_with_collection(
             )
 
         associated_data_sources.pop(data_source_fqn, None)
+
+        associated_data_sources = {
+            k: v.serialize() for k, v in associated_data_sources.items()
```
chiragjn, line 345: Same comment, why serialize instead of model_dump?

backend/types.py
S1LV3RJ1NX: done

chiragjn (approved):

PR #231, opened 2024-06-20
Remove Local MetadataStore and Update ReadMe

S1LV3RJ1NX:


README.md
```diff
@@ -396,11 +322,6 @@ Cognita makes it really easy to switch between parsers, loaders, models and retr
 
 -   Register the vectordb under `backend/modules/vector_db/__init__.py`
 
-### Rerankers:
```
chiragjn, line 399: Why remove this?

README.md
```diff
@@ -13,6 +13,7 @@ You can try out Cognita at: [https://cognita.truefoundry.com](https://cognita.tr
 
 # 🎉 What's new in Cognita
 
+-   [June, 2024] Cognita now supports it's own Metadatastore, powered by Prisma and Postgress. What it means for you is, you can now use Cognita via UI completely without the need of `local.metadata.yaml` file. You can create collections, data sources, and index them via UI. This makes it easier to use Cognita without any code changes.
```
chiragjn, line 16:
```suggestion
-   [June, 2024] Cognita now supports it's own Metadatastore, powered by Prisma and Postgress. You can now use Cognita via UI completely without the need of `local.metadata.yaml` file. You can create collections, data sources, and index them via UI. This makes it easier to use Cognita without any code changes.
```

README.md
S1LV3RJ1NX: Earlier there was a reranker file that handled the logic, now it's routed by infinity service. Can't change much here so removed it.

README.md
chiragjn: On another note, can we start creating releases? We can maybe use cal ver `2024.06.19_1`

We should be making one before major breaking changes so that people who want to continue using have a reference point
In this case the commit before merging this PR

chiragjn (approved):

PR #242, opened 2024-06-22
Fix models_config.json file issue

S1LV3RJ1NX:
Ignore registering models if API key is not present, and put out a war…ning. Also create a config.json file at runtime, helpful for template deployments

### DIFF
--- a/backend/modules/model_gateway/__init__.py
+++ b/backend/modules/model_gateway/__init__.py
@@ -1,1 +1,19 @@
+# When using locally, you can create your own models_config.json file before running docker-compose.
+# The following code is used for tf-deployments
+import json
+import os
 
+from backend.logger import logger
+
+# Check if models_config.sample.json exists but not models_config.json
+# If models_config.sample.json exists, copy it to models_config.json
+if os.path.exists("./models_config.sample.json") and not os.path.exists(
+    "./models_config.json"
+):
+    logger.info(
+        "models_config.json not found. Creating models_config.json from models_config.sample.json"
+    )
+    with open("./models_config.sample.json") as f:
+        data = json.load(f)
+        with open("./models_config.json", "w") as f:
+            json.dump(data, f, indent=4)
--- a/backend/modules/model_gateway/__init__.py
+++ b/backend/modules/model_gateway/__init__.py
@@ -4,16 +4,36 @@ import json
 import os
 
 from backend.logger import logger
+from backend.settings import settings
 
-# Check if models_config.sample.json exists but not models_config.json
-# If models_config.sample.json exists, copy it to models_config.json
-if os.path.exists("./models_config.sample.json") and not os.path.exists(
-    "./models_config.json"
+# Define the paths as constants
+# Current file's directory
+current_dir = os.path.dirname(__file__)
+
+# Navigate up three levels to get to the project root
+project_root = os.path.dirname(os.path.dirname(os.path.dirname(current_dir)))
+
+MODELS_CONFIG_SAMPLE_PATH = os.path.join(project_root, "models_config.sample.json")
+MODELS_CONFIG_PATH = os.path.join(project_root, "models_config.json")
+
+logger.info(f"MODELS_CONFIG_SAMPLE_PATH: {MODELS_CONFIG_SAMPLE_PATH}")
+logger.info(f"MODELS_CONFIG_PATH: {MODELS_CONFIG_PATH}")
+
+if (
+    settings.TFY_API_KEY
+    and os.path.exists(MODELS_CONFIG_SAMPLE_PATH)
+    and not os.path.exists(MODELS_CONFIG_PATH)
 ):
     logger.info(
         "models_config.json not found. Creating models_config.json from models_config.sample.json"
     )
-    with open("./models_config.sample.json") as f:
-        data = json.load(f)
-        with open("./models_config.json", "w") as f:
-            json.dump(data, f, indent=4)
+    data = {
+        "provider_name": "truefoundry",
+        "api_format": "openai",
+        "llm_model_ids": ["openai-main/gpt-4-turbo", "openai-main/gpt-3-5-turbo"],
+        "embedding_model_ids": ["openai-main/text-embedding-ada-002"],
+        "api_key_env_var": "TFY_API_KEY",
+        "base_url": settings.TFY_LLM_GATEWAY_URL,
+    }
+    with open(MODELS_CONFIG_PATH, "w") as f:
+        json.dump([data], f, indent=4)
--- a/backend/modules/model_gateway/model_gateway.py
+++ b/backend/modules/model_gateway/model_gateway.py
@@ -1,85 +1,119 @@
 import json
-from pydantic import BaseModel
-from typing import List, Optional
-from backend.types import ModelConfig, ModelProviderConfig, ModelType
-from langchain.embeddings.base import Embeddings
-from langchain_openai import OpenAIEmbeddings
-from langchain_core.language_models.chat_models import BaseChatModel
-from langchain_openai.chat_models import ChatOpenAI
 import os
+from typing import List
+
+from langchain.embeddings.base import Embeddings
+from langchain_core.language_models.chat_models import BaseChatModel
+from langchain_openai import OpenAIEmbeddings
+from langchain_openai.chat_models import ChatOpenAI
+
+from backend.logger import logger
+from backend.types import ModelConfig, ModelProviderConfig, ModelType
+
 
 class ModelGateway:
     config: List[ModelProviderConfig]
     modelsToProviderMap = {}
-    def __init__(self):        
+
+    def __init__(self):
         with open("./models_config.json") as f:
             data = json.load(f)
             self.config = [ModelProviderConfig(**item) for item in data]
             self.models: List[ModelConfig] = []
             for list in self.config:
-                if list.api_key_env_var and not os.environ.get(list.api_key_env_var):
-                    raise ValueError(f"Environment variable {list.api_key_env_var} not set.")
                 for model_id in list.embedding_model_ids:
-                    model_name = f"{list.provider_name}/{model_id}"
-                    self.modelsToProviderMap[model_name] = list
+                    if list.api_key_env_var and os.environ.get(list.api_key_env_var):
+                        # Register only those models whose API key is set
+                        model_name = f"{list.provider_name}/{model_id}"
+                        logger.info(f"Adding model {model_name} to the model gateway.")
+                        self.modelsToProviderMap[model_name] = list
+                    else:
+                        # Avoid adding the model to the map if the API key is not set
+                        logger.warning(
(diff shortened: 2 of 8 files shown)


backend/modules/model_gateway/__init__.py
```diff
@@ -1 +1,19 @@
...
+import json
+import os
 
+from backend.logger import logger
+
+# Check if models_config.sample.json exists but not models_config.json
+# If models_config.sample.json exists, copy it to models_config.json
+if os.path.exists("./models_config.sample.json") and not os.path.exists(
+    "./models_config.json"
+):
```
chiragjn, line 12: This can break depending on what is the cwd, the path should be made relative to this file using os.path operations on `__file__`

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
+from backend.types import ModelConfig, ModelProviderConfig, ModelType
+
 
 class ModelGateway:
     config: List[ModelProviderConfig]
     modelsToProviderMap = {}
-    def __init__(self):        
+
+    def __init__(self):
         with open("./models_config.json") as f:
```
chiragjn, line 19: Same comment about paths here

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
             data = json.load(f)
             self.config = [ModelProviderConfig(**item) for item in data]
             self.models: List[ModelConfig] = []
             for list in self.config:
```
chiragjn, line 23: We should not be using `list` as variable name as it shadows built in type name

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
 
 class ModelGateway:
     config: List[ModelProviderConfig]
     modelsToProviderMap = {}
-    def __init__(self):        
+
+    def __init__(self):
         with open("./models_config.json") as f:
             data = json.load(f)
             self.config = [ModelProviderConfig(**item) for item in data]
```
chiragjn, line 21: nit: this should have been `.parse_obj(item)`

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
-                    self.modelsToProviderMap[model_name] = list
+                    if list.api_key_env_var and os.environ.get(list.api_key_env_var):
+                        # Register only those models whose API key is set
+                        model_name = f"{list.provider_name}/{model_id}"
+                        logger.info(f"Adding model {model_name} to the model gateway.")
+                        self.modelsToProviderMap[model_name] = list
+                    else:
+                        # Avoid adding the model to the map if the API key is not set
+                        logger.warning(
+                            f"Environment variable {list.api_key_env_var} not set."
```
chiragjn, line 33: This error message should be more helpful
We should inform we are skipping adding the model to list of models

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
-            
+                    if list.api_key_env_var and os.environ.get(list.api_key_env_var):
+                        # Register only those models whose API key is set
+                        model_name = f"{list.provider_name}/{model_id}"
+                        logger.info(f"Adding model {model_name} to the model gateway.")
+                        self.modelsToProviderMap[model_name] = list
+                    else:
+                        # Avoid adding the model to the map if the API key is not set
+                        logger.warning(
+                            f"Environment variable {list.api_key_env_var} not set."
```
chiragjn, line 44: Same as above

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
 from langchain_openai.chat_models import ChatOpenAI
-import os
+
+from backend.logger import logger
+from backend.types import ModelConfig, ModelProviderConfig, ModelType
+
 
 class ModelGateway:
     config: List[ModelProviderConfig]
     modelsToProviderMap = {}
```
chiragjn, line 16: nit: convert to snake case
naming is also pretty confusing, isn't this `model_id_to_provider_config` ?

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
+                        self.modelsToProviderMap[model_name] = list
+                    else:
+                        # Avoid adding the model to the map if the API key is not set
+                        logger.warning(
+                            f"Environment variable {list.api_key_env_var} not set."
+                        )
+
     def get_embedding_models(self) -> List[ModelConfig]:
         models: List[ModelConfig] = []
         for list in self.config:
```
chiragjn, line 49: Why not compute this just once in `__init__`?

backend/modules/model_gateway/model_gateway.py
chiragjn, line 16: Do we get any benefit from making these static?

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
-                    type=ModelType.chat
-                ).dict())
-                              
+            if list.api_key_env_var and os.environ.get(list.api_key_env_var):
+                # Register only those models whose API key is set
+                for model_id in list.llm_model_ids:
+                    models.append(
+                        ModelConfig(
+                            name=f"{list.provider_name}/{model_id}", type=ModelType.chat
+                        ).dict()
```
chiragjn, line 71: Why `dict` ?
the return type says `List[ModelConfig]`

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
-                ).dict())
-                              
+            if list.api_key_env_var and os.environ.get(list.api_key_env_var):
+                # Register only those models whose API key is set
+                for model_id in list.embedding_model_ids:
+                    models.append(
+                        ModelConfig(
+                            name=f"{list.provider_name}/{model_id}",
+                            type=ModelType.embedding,
+                        ).dict()
```
chiragjn, line 57: Same here, why `.dict()` ?

backend/modules/model_gateway/model_gateway.py
```diff
@@ -1,85 +1,119 @@
...
-        model_provider_config: ModelProviderConfig = self.modelsToProviderMap[model_name]
+        model_provider_config: ModelProviderConfig = self.modelsToProviderMap[
+            model_name
+        ]
         if not model_provider_config.api_key_env_var:
             api_key = None
         else:
-            api_key = os.environ.get(model_provider_config.api_key_env_var, '')
+            api_key = os.environ.get(model_provider_config.api_key_env_var, "")
         model_id = "/".join(model_name.split("/")[1:])
```
chiragjn, line 86: We should be clear what is considered id and what is considered name
just from reading I would assume id is the fully qualified name but here is seems reverse

chiragjn, line 86:
```suggestion
        _, model_id = model_name.split("/", 1)
```

innoavator (changes requested):
We cannot merge this PR - we are just delaying the failure instead of asking the user to fix the issue. We should make the code independent of the model and power all the models via the gateway.

S1LV3RJ1NX:
> We cannot merge this PR - we are just delaying the failure instead of asking the user to fix the issue. We should make the code independent of the model and power all the models via the gateway.

The above changes were mostly done to allow the code to be executed on the platform too! One way I can see is, raise errors if models are not defined, and for TFY deployment purposes, generate the config.json at runtime. Does that work?

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/model_gateway/__init__.py
S1LV3RJ1NX: changed.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: @innoavator: Please check on this.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: Done.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: @innoavator: Please check on this.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: it was leading to JSON serialization error. Added a `to_dict()` to pydantic

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: @innoavator: Please check this

S1LV3RJ1NX:
> We cannot merge this PR - we are just delaying the failure instead of asking the user to fix the issue. We should make the code independent of the model and power all the models via the gateway.

@innoavator  I have now changed the config. User cannot proceed without setting up the config file correctly. Also fixed the config file generation issue while deployment. Please check some unresolved comments wrt to variable naming.

backend/types.py
```diff
@@ -104,6 +104,13 @@ class ModelConfig(BaseModel):
     type: Optional[ModelType]
     parameters: Optional[Dict[str, Any]] = None
 
+    def to_dict(self) -> Dict[str, Any]:
```
chiragjn, line 107: This doesn't seem to be used anywhere?

backend/server/routers/internal.py
```diff
@@ -113,6 +113,8 @@ def get_enabled_models(
             detail=f"Invalid model type: {model_type}",
         )
 
+    # Serialized models
+    serialized_models = [model.dict() for model in enabled_models]
     return JSONResponse(
-        content={"models": enabled_models},
+        content={"models": serialized_models},
```
chiragjn, line 119: This is fine for now, but normally we should define DTO Pydantic objects and just return the objects - fastapi will take care of serializing

innoavator (approved):

Issue #244, opened 2024-06-23
Error: P1001: Can't reach database server at `host.docker.internal`:`5432`


magick93:
Hello

I'm getting the following error when attempting to start using `docker compose --env-file compose.env up`:

```
cognita-backend   | Installing Prisma CLI
cognita-backend   |  * Install prebuilt node (22.3.0) ..... done.
cognita-backend   | Prisma schema loaded from backend/database/schema.prisma
cognita-backend   | Datasource "db": PostgreSQL database "cognita-config", schema "public" at "host.docker.internal:5432"
cognita-backend   | 
cognita-backend   | Error: P1001: Can't reach database server at `host.docker.internal`:`5432`
cognita-backend   | 
cognita-backend   | Please make sure your database server is running at `host.docker.internal`:`5432`.
Gracefully stopping... (press Ctrl+C again to force)
dependency failed to start: container cognita-backend is unhealthy
```

# Versions

- Docker Compose version v2.27.1

Client: Docker Engine - Community
 Version:           26.1.4
 API version:       1.44 (downgraded from 1.45)
 Go version:        go1.21.11
 Git commit:        5650f9b
 Built:             Wed Jun  5 11:28:57 2024
 OS/Arch:           linux/amd64
 Context:           rootless

Server: Docker Engine - Community
 Engine:
  Version:          25.0.2
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.21.6
  Git commit:       fce6e0c
  Built:            Thu Feb  1 00:23:03 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.28
  GitCommit:        ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
 rootlesskit:
  Version:          2.0.0
  ApiVersion:       1.1.1
  NetworkDriver:    slirp4netns
  PortDriver:       builtin
  StateDir:         /run/user/1000/dockerd-rootless
 slirp4netns:
  Version:          1.1.8
  GitCommit:        unknown

OS - Ubuntu 24.04 LTS

magick93:
If I modify the `compose.env` file to `DOCKER_HOST=db.cognita_cognita-docker` I'm able to pass the above error. 

However, I then get:

```
[+] Running 4/4
[...]

PR #245, opened 2024-06-24
Misc improvements and fixes for local setup

chiragjn:
- Fixed exception handling in prisma store
- Add prisma cli during build time
- Refactored providers in sample config
- Fix hosts and ports in compose.env
- Added a workflow for pre-commit


backend/Dockerfile
```diff
@@ -29,12 +29,9 @@ RUN if [ "${ADD_PARSER}" = "1" ]; then python3 -m pip install --use-pep517 --no-
 ARG ADD_VECTORDB=0
 RUN if [ "${ADD_VECTORDB}" = "1" ]; then python3 -m pip install --use-pep517 --no-cache-dir -r /tmp/vectordb.requirements.txt; fi
 
-# Temporary addition until templates have been updated using build args as environment variables
-ARG ADD_RERANKER_SVC_URL=""
```
S1LV3RJ1NX, line 33: Can we keep this, reranker is still required when deploying via templates, as there is no way to inject env vars at the moment.

compose.env
```diff
@@ -17,14 +17,14 @@ INFINITY_RERANKING_MODEL=mixedbread-ai/mxbai-rerank-xsmall-v1
...
-INFINITY_URL=http://infinity-server:7797
+INFINITY_URL=http://infinity-server:7997
 COGNITA_BACKEND_PORT=8000
 METADATA_STORE_CONFIG='{"provider":"prisma"}'
 VECTOR_DB_CONFIG='{"provider":"qdrant","url":"http://qdrant-server:6333", "config": {"grpc_port": 6334, "prefer_grpc": false}}'
 
 ## COGNITA_FRONTEND VARS
 COGNITA_FRONTEND_PORT=5001
-VITE_QA_FOUNDRY_URL=http://localhost:8000
+VITE_QA_FOUNDRY_URL=http://cognita-backend:8000
```
S1LV3RJ1NX, line 27: Front-end for some reason is not able to connect to `cognit-backend` hence had to add localhost.

docker-compose.yaml
```diff
@@ -128,9 +139,9 @@ services:
             - TFY_API_KEY=${TFY_API_KEY}
             - TFY_HOST=${TFY_HOST}
             - TFY_LLM_GATEWAY_URL=${TFY_LLM_GATEWAY_URL}
-            - DATABASE_URL=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:${POSTGRES_PORT}/cognita-config
+            - DATABASE_URL=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@cognita-db:5432/cognita-config
```
S1LV3RJ1NX, line 142: linux systems are not able to access db via service name. #244

backend/Dockerfile
chiragjn: Simply setting it in env var in TrueFoundry spec should work?

compose.env
chiragjn: Why? It should be able to connect, they are part of the same bridge network

I don't understand how will it work via localhost, unless frontend has requirement to talk to backend via host network.
Need to understand what breaks here

docker-compose.yaml
chiragjn: Why? docker dns is supposed to work for bridge network on any OS
We should reproduce and confirm

compose.env
chiragjn: I am able to use the UI just fine with these settings

S1LV3RJ1NX: Intresting, we can check on this... @sayan-truefoundry and I had issues.

backend/Dockerfile
chiragjn: I'll revert this but this is not the correct solution
Any change requires a rebuild

S1LV3RJ1NX (approved):

PR #247, opened 2024-06-24
Make models_config YAML for easy editing

chiragjn:


S1LV3RJ1NX (approved):

backend/modules/model_gateway/__init__.py
```diff
@@ -24,16 +26,18 @@
...
-    )
-    data = {
-        "provider_name": "truefoundry",
-        "api_format": "openai",
-        "llm_model_ids": ["openai-main/gpt-4-turbo", "openai-main/gpt-3-5-turbo"],
-        "embedding_model_ids": ["openai-main/text-embedding-ada-002"],
-        "api_key_env_var": "TFY_API_KEY",
-        "base_url": settings.TFY_LLM_GATEWAY_URL,
+    logger.info("models_config.yaml not found. Creating a default models_config.yaml")
+    config = {
```
innoavator, line 30: We shouldn't do this - this is very custom logic and shouldn't enter the main codebase. its better to ask the user to create a file or provide a default file.

innoavator (changes requested):

backend/modules/model_gateway/__init__.py
S1LV3RJ1NX: This is for only TF deployments.

backend/modules/model_gateway/__init__.py
```diff
@@ -1,10 +1,12 @@
 # When using locally, you can create your own models_config.json file before running docker-compose.
```
S1LV3RJ1NX, line 1: To rename this to models_config.yaml

S1LV3RJ1NX (changes requested):

innoavator (approved):

S1LV3RJ1NX (approved):

PR #252, opened 2024-06-24
Changes to Integrate SambaStudio LLM models and embedding models and replicate LLMs.

snova-kwasia:
1. Added some additional env vars to work with Sambastudio and replicate LLMs

2. Modified the get llm and get embedding model methods in model gateway to work with SambaStudio and Replicate, in the case that the model provider is one of the locals it will still default to ChatOpen ai. The main changes here are just branch logic to see if the provider is sambastudio or replicate and then instantiate those LLMs instead. As a result, I suggest returning an LLM object vs a BaseChat one as that will support more providers in the future.

3. Passed through a standard model provider parameters dict for model providers, this helps for cases in SambaStudio where different params are needed.

4. Updated the front end code to dynamically display the default params linked to each model provider based on which model the user has selected on the front end.


snova-kwasia:
@chiragjn - Love the work you folks are doing with cognita. Would be great to get a review of this and happy to talk to other folks about what we are doing.

innoavator (changes requested):
The method of adding models has completely been revamped. Please see the latest changes on main once. If the APIs are openai compatible, then no changes are needed.

snova-kwasia:
Thanks @innoavator for the response. I believe I'm using the latest main as of yesterday, with model gateway etc? What else has been revamped? I'm seeing the use of YAML vs JSON is this what you are referring to?

backend/modules/model_gateway/model_gateway.py
```diff
@@ -73,44 +66,79 @@ def get_embedder_from_model_config(self, model_name: str) -> Embeddings:
...
+            else None
         )
+        model_id = "/".join(model_config.name.split("/")[1:])
+
+        if model_provider_config.provider_name in [
+            "openai",
+            "local-ollama",
+            "local-infinity",
+            "truefoundry",
+        ]:
```
innoavator, line 114: @snova-kwasia : Does Samba studio not provide openai compatible APIs? If thats the case, it will be the easiest - else we should introduce it in the api_format flag - not whitelist the providers.

PR #253, opened 2024-06-24
Added reranking via model_gateway class

S1LV3RJ1NX:


models_config.truefoundry.yaml
```diff
@@ -1,10 +1,21 @@
...
+      api_key_env_var: TFY_API_KEY
+      llm_model_ids:
+          - "openai-main/gpt-4-turbo"
+          - "openai-main/gpt-3-5-turbo"
+      embedding_model_ids:
+          - "openai-main/text-embedding-ada-002"
+      reranking_model_ids: []
+
+    - provider_name: local-infinity
+      api_format: openai
```
innoavator, line 14: the api format is not correct? Shouldn't this be infinity?

innoavator (changes requested):

models_config.truefoundry.yaml
S1LV3RJ1NX: openai is used for embeddings

innoavator: Then there should be two entries in models_config json file.

innoavator (approved):

PR #258, opened 2024-06-27
Multimodal parser uses model gateway, replaced logger.error with logger.exception in try-catch

S1LV3RJ1NX:


backend/modules/parsers/multimodalparser.py
```diff
@@ -92,52 +108,41 @@ def __init__(
...
-                        "response": [
-                            page_number,
-                            result["choices"][0]["message"]["content"],
-                        ]
-                    }
-                if "error" in result:
-                    return {"error": result["error"]}
+        except Exception as e:
+            logger.exception(f"Error in page: {page_number} - {e}")
+            return {"error": e}
```
chiragjn, line 145: `e` may or may not be directly serializable

backend/modules/parsers/multimodalparser.py
```diff
@@ -92,52 +108,41 @@ def __init__(
...
-                                    "url": f"data:image/jpeg;base64,{base64_image}"
-                                },
-                            },
-                        ],
-                    }
-                ],
-                "max_tokens": 2048,
+                        "type": "image_url",
+                        "image_url": {
+                            "url": f"data:image/jpeg;base64,{base64_image}",
```
chiragjn, line 126: Are we sure, we'll always get jpeg?

backend/modules/parsers/multimodalparser.py
```diff
@@ -39,7 +43,22 @@ def arrayToBase64(image_arr: np.ndarray):
...
+        "parser_map": {
+            ".pdf": "MultiModalParser",
+        },
+        "additional_config": {
+            "model_configuration": {
+                # <provider>/<model_name> from model_config.yaml
+                "name": "truefoundry/openai-main/gpt-4-turbo"
+            }
+        }
+    }
```
chiragjn, line 61: This can get messy fast

I would prefer to design it like so

```
"parser_map": {
  ".pdf":  {
    "name": "MultiModalParser",
    "config": {
      // any arguments to pass to MultiModalParser init as kwargs
     }
}
```

chiragjn: Although I understand this would take more changes because of our existing structure

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: This works with png too. we just use base64 encoding of the image

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: Agree. We can add this as modification later. Where within config we have chunk size, chunk map and additional things. I'll add this as task

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: Resolved.

backend/modules/parsers/multimodalparser.py
```diff
@@ -182,7 +183,9 @@ async def get_chunks(
 
                     pages[page_number] = image_base64
                 except Exception as e:
-                    logger.error(f"Error in page: {page_number} - {e}")
+                    exc_info = sys.exc_info()
+                    logger.error("".join(traceback.format_exception(*exc_info)))
+                    logger.error(f"Error in page: {page_number}")
```
chiragjn, line 188: This is just replicating what `logger.exception` does
Might as well use that directly

backend/modules/parsers/multimodalparser.py
chiragjn: How was this resolved? I see the `return {"error": e}` is still as it is

chiragjn (approved):
Keeping it in approved state, but added a few comments

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: Got it, I actually wanted to have this in your serialisation comment.

chiragjn: This code is unrelated to that:
https://github.com/truefoundry/cognita/pull/258#discussion_r1660457670

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: The idea behind this was to mostly to skip pages that had errors but, yes, we can halt the parsing altogether.

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: Ohh, Then how to solve the serialization?

backend/modules/parsers/multimodalparser.py
chiragjn: Maybe this comment is not needed, I had a misunderstanding that this was returning this from get_chunks
I see the handling for this already exists at line 217

PR #264, opened 2024-07-04
Added Gateway logging for TF models

S1LV3RJ1NX:


chiragjn (approved):

models_config.truefoundry.yaml
```diff
@@ -9,6 +9,12 @@ model_providers:
       embedding_model_ids:
           - "openai-main/text-embedding-ada-002"
       reranking_model_ids: []
+      headers:
+          - "X-TFY-METADATA":
+                {
+                    "tfy_log_request": "true",
+                    "Custom-Metadata": "Cognita-LLM-Request",
+                }
```
chiragjn, line 17: Why is this a list?
this should just be `Dict[str, str]`

backend/types.py
```diff
@@ -121,6 +122,18 @@ class ModelProviderConfig(BaseModel):
     reranking_model_ids: List[str]
     api_key_env_var: str
     base_url: Optional[str] = None
+    headers: Optional[List] = None
```
chiragjn, line 125:
```suggestion
    headers: Dict[str, str] = Field(default_factory=dict)
```

backend/types.py
```diff
@@ -121,6 +122,18 @@ class ModelProviderConfig(BaseModel):
...
+    @property
+    def get_headers(self) -> Dict[str, str]:
+        formatted_headers = {}
+        for header in self.headers:
+            for key, value in header.items():
+                if isinstance(value, dict):
+                    formatted_headers[key] = json.dumps(value)
+                else:
+                    formatted_headers[key] = value
+        return formatted_headers
```
chiragjn, line 136: No need for this complexity in my opinion, let user's format in the yaml itself

models_config.truefoundry.yaml
S1LV3RJ1NX: Like how should the yaml format be?

backend/types.py
S1LV3RJ1NX: gateway requires dict to be json.dumps()

chiragjn (approved):

PR #283, opened 2024-07-20
Fix backend Dockerfile

falmanna:
remove `parsers.requirements.txt`


S1LV3RJ1NX:
Hi @falmanna , thanks for the update.
The above change is being fixed in another major PR coming up soon, hence wont be approving this.

S1LV3RJ1NX (approved):

PR #286, opened 2024-07-23
[Breaking Change] Refactor parser configuration, embedding configuration and misc fixes

S1LV3RJ1NX:


compose.env
```diff
@@ -17,7 +17,7 @@ UNSTRUCTURED_IO_URL=http://unstructured-io-parsers:9500/
 
 ## COGNITA_FRONTEND VARS
 COGNITA_FRONTEND_PORT=5001
-VITE_QA_FOUNDRY_URL=http://localhost:8000
+VITE_QA_FOUNDRY_URL=http://cognita-backend:8000
```
chiragjn, line 20: This is probably not correct, localhost was correct, because the host browser cannot resolve `cognita-backend`

compose.env
```diff
@@ -32,8 +32,6 @@ OLLAMA_MODEL=qwen2:1.5b
 ## INFINITY VARS
 INFINITY_EMBEDDING_MODEL=mixedbread-ai/mxbai-embed-large-v1
 INFINITY_RERANKING_MODEL=mixedbread-ai/mxbai-rerank-xsmall-v1
-## INFINITY_API_KEY, only required if you enable API KEY auth on infinity container
-INFINITY_API_KEY=
```
chiragjn, line 36: Any problems in keeping this with empty value?

frontend/Dockerfile.dev
chiragjn, line 1: Instead of a whole new build file, based on a environment variable/build arg
we should be able 
- decide if we want to build dist or not
- do `yarn serve:dev` and `yarn serve` using `package.json`


This doesn't necessarily have to be in this PR

S1LV3RJ1NX: Yep, I wanted to do this, but was not sure.

compose.env
S1LV3RJ1NX: works we can keep an empty API key.

compose.env
S1LV3RJ1NX: Will do this change,

backend/modules/dataloaders/truefoundryloader.py
```diff
@@ -82,6 +82,8 @@ def load_filtered_data(
                     if f.startswith("."):
                         continue
                     full_path = os.path.join(root, f)
+                    if ".truefoundry" in full_path:
+                        continue
```
chiragjn, line 86: is there a case where we would want to iterate over dest_dir instead of just `download_info`

(`download_info` should have been called `artifact_files_dir`)

Then you would not need this check

S1LV3RJ1NX: Can call `download_info` as `artifact_files_dir` but then there is `data_dir` as well, what might be a suitable name.
You are write, can walk over download info rather than dest_dir. Will make that change.

models_config.sample.yaml
chiragjn, line 1: What has changed in this file?

models_config.sample.yaml
```diff
@@ -4,57 +4,56 @@ model_providers:
...
+      llm_model_ids:
+          - "qwen2:1.5b"
+      embedding_model_ids: []
+      reranking_model_ids: []
+      default_headers: {}
 
+    - provider_name: local-infinity
+      api_format: openai
+      base_url: http://infinity-server:7997/
+      api_key_env_var: INFINITY_API_KEY
```
chiragjn, line 20: Is this required to set?

models_config.sample.yaml
```diff
@@ -4,57 +4,56 @@ model_providers:
...
-  #   default_headers: {}
+# - provider_name: truefoundry
+#   api_format: openai
+#   base_url: https://llm-gateway.truefoundry.com/api/inference/openai
+#   api_key_env_var: TFY_API_KEY
+#   llm_model_ids:
+#     - "openai-main/gpt-4-turbo"
+#     - "openai-main/gpt-3-5-turbo"
+#   embedding_model_ids:
+#     - "openai-main/text-embedding-ada-002"
```
chiragjn, line 57: minor add gpt-4o-mini and small embedding model ids

models_config.truefoundry.yaml
chiragjn, line 1: nitpick: Normally we indent all yaml files with 2 spaces

backend/modules/parsers/unstructured_io.py
```diff
@@ -36,11 +36,11 @@ class UnstructuredIoParser(BaseParser):
         ".xlsx",
     ]
 
-    def __init__(self, max_chunk_size: int = 2000, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
```
chiragjn, line 39:
```suggestion
    def __init__(self, *, max_chunk_size: int = 2000, **kwargs):
```

backend/modules/parsers/unstructured_io.py
```diff
@@ -36,11 +36,11 @@ class UnstructuredIoParser(BaseParser):
...
         """
         Initializes the UnstructuredIoParser object.
         """
-        self.max_chunk_size = max_chunk_size
+        self.max_chunk_size = kwargs.get("max_chunk_size", 2000)
```
chiragjn, line 43:
```suggestion
        self.max_chunk_size = max_chunk_size
```

backend/modules/parsers/unstructured_io.py
chiragjn: Ideally we should create change the base class of Parsers to pydantic BaseModel

In the future we want to iterate over all registered parsers, get their openapi / json schema so that we can show full config on UI

Only possible challenge I see is how we declare private fields with pydantic
Maybe Pydantic v2 has better solution for this.

---

Another alternative can be

```
class BaseParser(abc.ABC):
    class InitArgs(BaseModel):
        pass
    def __init__(self, args: InitArgs):
        ...


class UnstructuredIoParser(BaseParser):
    class InitKwargs(BaseParser.InitArgs):
        max_chunk_size: int = 2000
    def __init__(self, args: InitArgs):
        ...
```

I don't like the extra toil in unwrapping `args` but seems like a decent tradeoff

> class names can be changed

chiragjn: https://docs.pydantic.dev/latest/concepts/models/#private-model-attributes

v2 does have private attrs, just need to make the field name start with `_`

docker-compose.yaml
```diff
@@ -71,6 +71,7 @@ services:
         environment:
             - INFINITY_MODEL_ID=${INFINITY_EMBEDDING_MODEL};${INFINITY_RERANKING_MODEL}
             - INFINITY_BATCH_SIZE=8
+            - API_KEY=${INFINITY_API_KEY}
```
chiragjn, line 74: This seems incorrect?

```suggestion
            - INFINITY_API_KEY=${INFINITY_API_KEY}
```

backend/indexer/types.py
```diff
@@ -20,7 +22,7 @@ class DataIngestionConfig(BaseModel):
     embedder_config: EmbedderConfig = Field(
         title="Embedder configuration",
     )
-    parser_config: ParserConfig = Field(
+    parser_config: Dict[str, Union[ParserConfig, str]] = Field(
```
chiragjn, line 25: What happens when I give `str` as value?

I was wondering we would just do
```suggestion
    parser_config: Dict[str, ParserConfig] = Field(
```

backend/modules/query_controllers/multimodal/types.py
```diff
@@ -155,3 +156,13 @@ def validate_retriever_type(cls, values: Dict) -> Dict:
...
+
+
+class Answer(BaseModel):
+    type: str = "answer"
+    content: str
+
+
+class Docs(BaseModel):
+    type: str = "docs"
+    content: Optional[List[Document]] = None
```
chiragjn, line 168: Is nullability important?

```suggestion
    content: List[Document] = Field(default_factory=list)
```

backend/types.py
```diff
@@ -125,32 +125,22 @@ class ModelProviderConfig(BaseModel):
     default_headers: Dict[str, str] = Field(default_factory=dict)
 
 
-class EmbedderConfig(BaseModel):
+class EmbedderConfig(ModelConfig):
```
chiragjn, line 128: I would recommend follow the same structure as parser?

```
embedder: str
kwargs: Dict[str, Any]
```

chiragjn: or we make parser config also look similar

```
type: parser
name: MultiModalParser
parameters: { ... }
```

backend/types.py
```diff
@@ -257,7 +247,7 @@ class BaseDataIngestionRun(BaseModel):
         title="Fully qualified name of the data source",
     )
 
-    parser_config: ParserConfig = Field(
+    parser_config: Dict[str, Union[ParserConfig, str]] = Field(
```
chiragjn, line 250: For all instances, check if we need str in values

backend/modules/parsers/multimodalparser.py
```diff
@@ -61,29 +61,27 @@ class MultiModalParser(BaseParser):
 
     def __init__(
         self,
-        additional_config: dict = None,
         *args,
         **kwargs,
```
chiragjn, line 65: same comment as here: https://github.com/truefoundry/cognita/pull/286/files?show-viewed-files=true&file-filters%5B%5D=#r1693115703

backend/modules/query_controllers/multimodal/controller.py
```diff
@@ -180,38 +201,31 @@ async def _get_retriever(self, vector_store, retriever_name, retriever_config):
...
+            formatted_docs.append(
+                Document(page_content=doc.page_content, metadata=metadata)
+            )
+        return formatted_docs
+
+    async def _sse_wrap(self, gen):
+        async for data in gen:
+            yield "event: data\n"
+            yield f"data: {json.dumps(data.dict())}\n\n"
+        yield "event: end\n"
```
chiragjn, line 221: Let's make this part of Base?

backend/modules/parsers/parser.py
```diff
@@ -73,25 +76,20 @@ def get_parser_for_extension(
     if file_extension not in parsers_map:
         # get the first parser name registered with the extension
         name = PARSER_REGISTRY_EXTENSIONS[file_extension][0]
-        print(
-            f"Parser map not found in the collection for extension {file_extension}. Hence, using parser {name}"
-        )
         logger.debug(
             f"Parser map not found in the collection for extension {file_extension}. Hence, using parser {name}"
         )
+        return PARSER_REGISTRY[name](*args, **kwargs)
```
chiragjn, line 82: We should avoid `*args` wherever we can

backend/modules/query_controllers/example/controller.py
```diff
@@ -180,19 +188,33 @@ async def _get_retriever(self, vector_store, retriever_name, retriever_config):
...
+                key: doc.metadata[key]
+                for key in self.required_metadata
+                if key in doc.metadata
+            }
+            formatted_docs.append(
+                Document(page_content=doc.page_content, metadata=metadata)
+            )
+        return formatted_docs
+
+    async def _sse_wrap(self, gen):
```
chiragjn, line 204: Will move to base class

chiragjn (commented):
Mostly fine, added some suggestions

docker-compose.yaml
S1LV3RJ1NX: Nope, Infinity service has API_KEY as env variable.

backend/indexer/types.py
S1LV3RJ1NX: This was for backward compatiblity, but now I'll be removing it.

backend/modules/query_controllers/example/types.py
```diff
@@ -155,3 +156,13 @@ def validate_retriever_type(cls, values: Dict) -> Dict:
...
+
+
+class Answer(BaseModel):
+    type: str = "answer"
+    content: str
+
+
+class Docs(BaseModel):
+    type: str = "docs"
+    content: Optional[List[Document]] = None
```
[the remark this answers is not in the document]
S1LV3RJ1NX: Nope. Doing the change.

backend/modules/query_controllers/multimodal/controller.py
S1LV3RJ1NX: We do not have a Base class yet, need to this as refactor to push common functionalities to Base.

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: The link does not open but if the comment is regarding max_chunk_size, this is not used in `MultiModalParser`

backend/database/schema.prisma
```diff
@@ -24,7 +24,7 @@ model Collection {
     description             String?
     embedder_config         Json
     // Collection can have multiple data sources
-    associated_data_sources Json?
+    associated_data_sources Json    @default("{}")
```
chiragjn, line 27: Will primsa auto migrate older NULL entries?

backend/modules/metadata_store/prismastore.py
```diff
@@ -447,7 +447,7 @@ async def aget_data_ingestion_runs(
         """Get all data ingestion runs for a collection"""
         try:
             data_ingestion_runs = await self.db.ingestionruns.find_many(
-                where={"collection_name": collection_name}
+                where={"collection_name": collection_name}, order={"id": "desc"}
```
chiragjn, line 450: Should also add this Collection listing APIs?

backend/modules/parsers/multimodalparser.py
```diff
@@ -59,29 +59,23 @@ class MultiModalParser(BaseParser):
 
     supported_file_extensions = [".pdf", ".png", ".jpeg", ".jpg"]
 
-    def __init__(
-        self,
-        *args,
-        **kwargs,
-    ):
+    def __init__(self, *, model_configuration: dict = {}, prompt: str = "", **kwargs):
```
chiragjn, line 62:
```suggestion
    def __init__(self, *, model_configuration: ModelConfig = None, prompt: str = "", **kwargs):
```

1. We should not use mutable defaults
2. Let the type indicate we want `ModelConfig` instance


Line 69 will ensure we have converted to ModelConfig even if get a dict

chiragjn: This is one of the reasons why I want to move base classes to pydantic

so pydantic can take care of casting and conversion and we get schema for free

backend/modules/parsers/multimodalparser.py
```diff
@@ -94,8 +88,6 @@ def __init__(
 Comparisons: Compare different charts within the same image or compare data points within a single chart. Highlight similarities, differences, or correlations between datasets.
 Conclude with a summary of the key findings from your analysis and any recommendations based on those findings."""
 
-            super().__init__(*args, **kwargs)
-
```
chiragjn, line 100: Why remove this?

backend/modules/query_controllers/example/types.py
```diff
@@ -165,4 +165,4 @@ class Answer(BaseModel):
 
 class Docs(BaseModel):
     type: str = "docs"
-    content: Optional[List[Document]] = None
+    content: Optional[List[Document]] = Field(default_factory=list)
```
chiragjn, line 168: maybe drop Optional

```suggestion
    content: List[Document] = Field(default_factory=list)
```

backend/modules/query_controllers/multimodal/types.py
```diff
@@ -165,4 +165,4 @@ class Answer(BaseModel):
 
 class Docs(BaseModel):
     type: str = "docs"
-    content: Optional[List[Document]] = None
+    content: Optional[List[Document]] = Field(default_factory=list)
```
chiragjn, line 168: maybe drop Optional

```suggestion
    content: List[Document] = Field(default_factory=list)
```

backend/types.py
```diff
@@ -133,15 +134,13 @@ class EmbedderConfig(ModelConfig):
...
     """
     Parser configuration
     """
 
-    parser: str = Field(title="Parser Class Name")
-    kwargs: Optional[Dict[str, Any]] = Field(
-        title="Parser Class Arguments", default_factory=dict
-    )
+    type: ModelType = ModelType.parser
+    pass
```
chiragjn, line 143: Uh, I would rather just create another model and enum called ModuleConfig and ModuleType with the same structure

ModelConfig can envolve in future with model specific things

chiragjn: Kinda debating where does Embedder fit?
Embedder itself is a module but emedding models should be Model

Maybe then we reserve ModelConfig for model objects in the gateway class

backend/modules/parsers/unstructured_io.py
chiragjn: Keeping this open for future reference

chiragjn (approved):
Added a few comments
Keeping this in approved state

backend/database/schema.prisma
S1LV3RJ1NX: There are no null values in the DB. It's only during collection creation API is split first into creating the collection and attaching the DB. But yes if there are prisma should be able to do.

backend/modules/metadata_store/prismastore.py
S1LV3RJ1NX: Sure, will do. So that newer collections are on the top.

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: Sure.

backend/modules/parsers/multimodalparser.py
S1LV3RJ1NX: Was not sure, why was this in the first place, not required right? Didn't find any specific reason where we are using BaseParser directly.

chiragjn: Generally, `super().__init__(...)` should be called for correctness in case base class has some logic to finish common init logic

S1LV3RJ1NX: Will revert back.

backend/types.py
S1LV3RJ1NX: For now, I'll keep it as is in the PR, maybe we can decide in some other PR on what do do.

PR #308, opened 2024-08-23
Azure bedrock support

dorhunova-n-ix:


dorhunova:
pls ignore, mistakenly pushed here

PR #312, opened 2024-08-27
Implement enter to submit

Blakeinstein:
This was a pain point I had while using the UI. I have added a quick fix to submit input when typing and pressing enter.

Another requested feature was shift+enter for a new line, but that would be a more significant change:
1. Requires changing `input` to `textarea`
2. Handling a custom event where enter means submit instead.
3. Changing height for the input field.


chiragjn (approved):
@sayan-truefoundry would be best to check this, but approving because seems like a good change

sayan-truefoundry (approved):

PR #320, opened 2024-09-05
Fix unassociate datasrc api of prisma store

S1LV3RJ1NX:


backend/modules/metadata_store/prismastore.py
```diff
@@ -340,17 +340,26 @@ async def aunassociate_data_source_with_collection(
                 status_code=400,
                 detail=f"Data source with fqn {data_source_fqn} not associated with collection {collection_name}",
             )
-
         associated_data_sources.pop(data_source_fqn, None)
```
chiragjn, line 343: It is quite confusing what is the type of associated_data_sources?

Line 324 annotates it as pydantic object, what exactly is returned from DB?

S1LV3RJ1NX: Actually it does not it is of form dict: dict

S1LV3RJ1NX: I have removed the type annotation. Actually from DB we get of form str: AsociatedDataSrcs

chiragjn (approved):
Keeping this approved, but we should check why can't we get correct data types from prisma itself

Issue #319, opened 2024-09-05
Run the indexer as a background job when running in local Cognita mode


innoavator:
Currently when you create a collection locally, indexing is triggered automatically by frontend and it runs in the same process. As a result, collection creation blocks all other API calls
Hence we would want indexing to start as a separate background process that does not block the collection creation.

Starting point is:
- ingest api in `backend/server/routers/collection.py` and  trace back the necessary files from there.

PR #321, opened 2024-09-05
Feature: background ingestion in LOCAL mode

kunwar31:
Solves https://github.com/truefoundry/cognita/issues/319


backend/indexer/indexer.py
```diff
@@ -63,11 +67,15 @@ async def sync_data_source_to_collection(inputs: DataIngestionConfig):
     Returns:
         None
     """
-    client = await get_client()
-    await client.aupdate_data_ingestion_run_status(
+    loop = asyncio.new_event_loop()
+    def _run_async(coroutine):
+        return loop.run_until_complete(coroutine)
```
chiragjn, line 72: We should not need a new loop, fastapi already runs an event loop

backend/indexer/indexer.py
```diff
@@ -82,54 +90,55 @@ async def sync_data_source_to_collection(inputs: DataIngestionConfig):
         )
     except Exception as e:
         logger.exception(e)
-        await client.aupdate_data_ingestion_run_status(
+        _run_async( client.aupdate_data_ingestion_run_status(
```
chiragjn, line 93: We need not change await calls, they already run on the same event loop managed by uvicorn and starlette

backend/indexer/indexer.py
```diff
@@ -42,17 +44,19 @@ def get_data_point_fqn_to_hash_map(
     return data_point_fqn_to_hash
 
 
-async def sync_data_source_to_collection(inputs: DataIngestionConfig):
+def sync_data_source_to_collection(inputs: DataIngestionConfig):
```
chiragjn, line 47: Should not change this

backend/indexer/indexer.py
```diff
@@ -364,8 +372,8 @@ async def ingest_data(request: IngestDataToCollectionDto):
                         raise_error_on_failure=created_data_ingestion_run.raise_error_on_failure,
                         batch_size=request.batch_size,
                     )
-                )
-                created_data_ingestion_run.status = DataIngestionRunStatus.COMPLETED
+                Process(target=sync_data_source_to_collection, args=(ingestion_config, )).start()
```
chiragjn, line 375: While this can work, there is no manager for this process and can run amok.
It would be better to rather use starlette's run_in_threadpool to start with

chiragjn: or current loop's run_in_executor with a ProcessPoolExecutor

chiragjn: However both of these will require to await and block the request from finishing.
So, it might be better to start with 
https://fastapi.tiangolo.com/tutorial/background-tasks/
https://www.starlette.io/background/

Long term we want to move to proper background workers Celery/Dramatiq/Temporal/TrueFoundry Asycn Service

backend/indexer/indexer.py
kunwar31: Running the ingestion task in fastapi's event loop is the reason why it blocks the entire API

backend/indexer/indexer.py
kunwar31: Even if I use background task of fastAPI, it just awaits async functions (see [this](https://github.com/encode/starlette/blob/master/starlette/background.py#L25)), which will block fastapi's event loop

backend/indexer/indexer.py
kunwar31: @chiragjn looking at the source for background tasks, it will just await the task on the same event loop, even if we pass a synchronous function, it runs it in a thread pool and "awaits" completion

kunwar31: @chiragjn other people have had similar challenge with BackgroundTasks. https://stackoverflow.com/questions/76328904/is-there-a-difference-between-starlette-fastapi-background-tasks-and-simply-usin

kunwar31:
@chiragjn The main issue is that there is no "manager" for the process i created, and I have intentionally not added it, as this is only meant for LOCAL ingestions. 
1. I can add another deamon thread to manage any running background process, which can update the status as FAILED if the process dies unexpectedly
2. However, the above solution will again add futher unwanted complexity. I agree that long term we need a proper solution (based on queues + background workers)

backend/indexer/indexer.py
chiragjn: Then simplest way is to add a global ProcessPoolExecutor with fixed set of n workers - create it in the lifespan context manager of fastapi and terminate/shutdown it in the same lifespan context manager

Submit to the pool and return immediately with `201` no need to wait for indexing to finish, the status will be updated anyway via db

kunwar31:
@chiragjn changed to use ProcessPoolExecutor, initialized in the lifespan context manager of fastAPI.
number of workers can be set using PROCESS_POOL_WORKERS environment variable (have set default to 4 in compose.env, 1 in Settings class)
Now, before fastAPI shuts down it will ensure the there are no tasks running in the process pool.
However, unexpected termination of ingestion process will still not be tracked

chiragjn:
However, unexpected termination of ingestion process will still not be tracked
That's alright, we have to start somewhere :)

backend/indexer/indexer.py
chiragjn: Let's please revert changes to this function

backend/indexer/indexer.py
```diff
@@ -364,8 +371,18 @@ async def ingest_data(request: IngestDataToCollectionDto):
                         raise_error_on_failure=created_data_ingestion_run.raise_error_on_failure,
                         batch_size=request.batch_size,
                     )
-                )
-                created_data_ingestion_run.status = DataIngestionRunStatus.COMPLETED
+                # Import here to avoid circular dependency
+                from backend.server.app import process_pool
+                # future of this submission is ignored, ingestion failure due to process termination will not be tracked
```
chiragjn, line 376: Let's not cross import
it is fine to pass `pool` as input arg which is by default `None`

chiragjn: We want to keep all the code as is and rather use something on lines of 

```
loop = asyncio.get_running_loop()
asyncio.create_task(loop.run_in_executor(sync_data_source_to_collection(...), pool, ...))
```

kunwar31: @chiragjn [run_in_executor](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) is  used to run synchronous code, so sync_data_source_to_collection can't be async
further, asyncio.create_task(coro) will again block the event loop (waiting for coro to finish at some point)

chiragjn: > [run_in_executor](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) is used to run synchronous code

You are correct, I am not framing it correctly, give me some time, I'll get back on this

kunwar31: @chiragjn I spend a few hours thinking about it too, I know you want to run the whole ingestion in the same async event loop, but there are blocking synchronous parts in the sync_data_source_to_collection function, mainly the getting of chunks from unstructured_api, which will keep blocking other tasks in the event loop.
   
   this method get_chunks in 
   backend.modules.parsers.unstructured_io.UnstructuredIoParser
   does a very long POST call which blocks everything
   
if we make somehow make this request non-blocking, we would be in a better shape, but there still wont be any guarantee that it doesn't block the main event loop.

kunwar31: > Let's not cross import it is fine to pass `pool` as input arg which is by default `None`

@chiragjn add a process_pool module to make this cleaner.

chiragjn: > I know you want to run the whole ingestion in the same async event loop,

Actually I don't want to in run in same loop. I want to run stuff in a separate process, my only hesistation is wrapping all async coroutines to be sync again, just trying to see if we can somehow avoid that. Worst case we have to take this route

> this method get_chunks in
> backend.modules.parsers.unstructured_io.UnstructuredIoParser
> does a very long POST call which blocks everything
>
> if we make somehow make this request non-blocking, we would be in a better shape, but there still wont be any guarantee that it doesn't block the main event loop.

Yes this and plus the whole vector store implementation also needs to be made async.
At least the unstructured API call should be easy to change, instead of requests we just have to use asyncio httpx or  aiohttp

> but there still wont be any guarantee that it doesn't block the main event loop.

That is true, some bad parser can still block, hence eventually we will move away to a queue and workers system.

kunwar31:
> Actually I don't want to in run in same loop. I want to run stuff in a separate process, my only hesistation is wrapping all async coroutines to be sync again, just trying to see if we can somehow avoid that. Worst case we have to wrap and just accept it

@chiragjn I think i've understood what you meant. I have simplified the diff, created a generic `AsyncProcessPoolExecutor` in `process_pool` module which handles all the steps. no need to wrap individual functions

backend/indexer/indexer.py
```diff
@@ -364,8 +365,9 @@ async def ingest_data(request: IngestDataToCollectionDto):
                         raise_error_on_failure=created_data_ingestion_run.raise_error_on_failure,
                         batch_size=request.batch_size,
                     )
-                )
-                created_data_ingestion_run.status = DataIngestionRunStatus.COMPLETED
+                # future of this submission is ignored, ingestion failure due to process termination will not be tracked
+                process_pool.submit(sync_data_source_to_collection, ingestion_config)
```
chiragjn, line 369: This is almost what I was going for!


Here is what I want to get to https://gist.github.com/chiragjn/882d43a83583a9729a61ee25e2b60c5c

I suppose we will have to wrap `sync_data_source_to_collection` in `asyncio.run` on the process worker to run it, otherwise it won't schedule

chiragjn: Ah I see you ended up doing this wrapping in the custom pool implementation, nice

backend/modules/process_pool/__init__.py
chiragjn, line 1: doesn't have to be in modules
Normally modules for this repo are re-usable components in RAG pipelines

This can be just in server utils

backend/modules/process_pool/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+
+from backend.settings import settings
+
+
+class AsyncProcessPoolExecutor(ProcessPoolExecutor):
+    @staticmethod
+    def _async_to_sync(__fn, *args, **kwarg):
+        loop = asyncio.new_event_loop()
+        result = loop.run_until_complete(__fn(*args, **kwarg))
+        loop.close()
```
chiragjn, line 15: just `asyncio.run(__fn(*args, **kwargs))` should be good I guess if we only expect to run one coroutine at a time

backend/modules/process_pool/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from concurrent.futures import ProcessPoolExecutor
+
+from fastapi import FastAPI
+
+from backend.settings import settings
+
+
+class AsyncProcessPoolExecutor(ProcessPoolExecutor):
+    @staticmethod
+    def _async_to_sync(__fn, *args, **kwarg):
```
chiragjn, line 12:
```suggestion
    def _async_to_sync(__fn, *args, **kwargs):
```

backend/indexer/indexer.py
chiragjn: Let's just pass the `process_pool` as optional arg

backend/modules/process_pool/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+        return super().submit(self._async_to_sync, __fn, *args, **kwargs)
+
+
+process_pool = AsyncProcessPoolExecutor(max_workers=settings.PROCESS_POOL_WORKERS)
+
+
+@asynccontextmanager
+async def process_pool_lifespan_manager(app: FastAPI):
+    global process_pool
+    yield  # FastAPI runs here
```
chiragjn, line 28: don't need it to global, can just bind it to `app.state`

chiragjn:
We are almost there, I like this version of the code, just some code organizing changes. Thanks for being really patient and iterating with us!

backend/modules/process_pool/__init__.py
kunwar31: _async_to_sync can be replaced with asyncio.run, yes. I think it does the same thing under the hood, when using it with ProcessPoolExecutor. For threads, it re-uses the existing event loop

kunwar31:
> We are almost there, I like this version of the code, just some code organizing changes. Thanks for being really patient and iterating with us!

@chiragjn Refactored 
1. AsyncProcessPoolExecutor is now in utils.py
2. api.state contains the process_pool, which is passed to ingestion worker as an arg
3. had to "hack" around in router's ingest method to get the app.state, not ideal

kunwar31:
Tested this by running 4 ingestion jobs, 2 data sources and 2 collections, all 4 run in background parallely (workers=4) and backend is responsive

chiragjn (approved):
🎉

PR #323, opened 2024-09-06
Carbon dataloader and Internet search feature 

S1LV3RJ1NX:


backend/modules/dataloaders/carbondataloader.py
```diff
@@ -0,0 +1,187 @@
...
+
+        while total == -1 or count < total:
+            response = self._request(
+                "POST", "https://api.carbon.ai/user_files_v2", json=payload
+            )
+            page = _UserFilesV2Response.parse_obj(response)
+            if total == -1:
+                total = page.count
+            for file in page.results:
+                # TODO (chiragjn): There can be an edge case here where file.file_metadata.is_folder = True
```
chiragjn, line 85: I never got around to this.
Actually now that we support pydantic v2, we should be able to use the official carbon python sdk

chiragjn (approved):
Approving to unblock

backend/modules/dataloaders/carbondataloader.py
```diff
@@ -0,0 +1,187 @@
...
+            payload["include_additional_files"] = include_additional_files
+
+        total = -1
+        count = 0
+
+        while total == -1 or count < total:
+            response = self._request(
+                "POST", "https://api.carbon.ai/user_files_v2", json=payload
+            )
+            page = _UserFilesV2Response.parse_obj(response)
```
chiragjn, line 81: This will break actually, this needs to be `.model_validate`

backend/modules/query_controllers/example/controller.py
```diff
@@ -32,6 +32,7 @@ async def answer(
         """
         Sample answer method to answer the question using the context from the collection
         """
+        logger.info(f"Request: {request.dict()}")
```
chiragjn, line 35: `.model_dump()`

PR #325, opened 2024-09-06
Removed TF Metadatastore and necessary changes

S1LV3RJ1NX:
Template change: https://github.com/truefoundry/truefoundry-templates/pull/110


backend/modules/metadata_store/base.py
```diff
@@ -18,60 +18,29 @@
 from backend.utils import run_in_executor
 
 
-# TODO (chiragjn): Ideal would be we make `async def a*` abstract methods and drop sync ones
-#   Implementations can then opt to call their sync versions using run_in_executor
 class BaseMetadataStore(ABC):
```
chiragjn, line 21: All the not implemented methods which are required should be marked with `@abstractmethod`

frontend/Dockerfile
```diff
@@ -6,29 +6,27 @@ COPY . .
...
-    fi
-RUN if [ -n "$VITE_USE_LOCAL" ]; then \
-    export VITE_USE_LOCAL=${VITE_USE_LOCAL}; \
-    fi
-RUN if [ -n "$VITE_GTAG_ID" ]; then \
-    export VITE_GTAG_ID=${VITE_GTAG_ID}; \
-    fi
-RUN if [ -n "$VITE_CARBON_API_KEY" ]; then \
-    export VITE_CARBON_API_KEY=${VITE_CARBON_API_KEY}; \
-    fi
```
chiragjn, line 31: Why remove the conditionals here? What if these are provided in build args, do we still get the correct behavior?

frontend/Dockerfile
```diff
@@ -6,29 +6,27 @@ COPY . .
 # Install dependencies
 RUN yarn install --frozen-lockfile
 
-# Copy environment variables file if it doesn't exist
-RUN test -e .env || cp .env.example .env
```
chiragjn, line 10: If we are removing this, then do we need to define defaults for each ARG

models_config.custom.yaml
chiragjn, line 1: Do we really need to add this file?

.github/workflows/tfy-deploy.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - ci-cd
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Install truefoundry
+        run: pip install -U --pre truefoundry
```
chiragjn, line 15:
```suggestion
        run: pip install -U truefoundry
```

truefoundry.yaml
```diff
@@ -0,0 +1,403 @@
...
+      BATCH_SIZE: "4"
+    name: cas-infinity
+    type: service
+    image:
+      type: image
+      command: >-
+        infinity_emb v2 --model-id mixedbread-ai/mxbai-embed-large-v1 --model-id
+        mixedbread-ai/mxbai-rerank-xsmall-v1 --port $(PORT) --batch-size
+        $(BATCH_SIZE) --api-key $(API_KEY)
+      image_uri: michaelf34/infinity:0.0.50
```
chiragjn, line 356:
```suggestion
      image_uri: michaelf34/infinity:0.0.54
```

truefoundry.yaml
```diff
@@ -0,0 +1,403 @@
...
+
+  # Qdrant-UI
+  - name: cas-qdrant-ui
+    type: service
+    image:
+      type: build
+      build_source:
+        type: git
+        repo_url: https://github.com/truefoundry/qdrant-web-ui-new
+        branch_name: support-path-based-routing
```
chiragjn, line 291: Last I remember there was some header based hack we had to do to make path based routing work. 
I think in general it would be better we use `path` for all services, that way we ensure path based routing is always tested

frontend/Dockerfile
S1LV3RJ1NX: yes they work

models_config.custom.yaml
S1LV3RJ1NX: ohh yes, will remove this and replace with `models_config.truefoundry.yaml`

frontend/Dockerfile
S1LV3RJ1NX: yes we can remove that.

innoavator (approved):

.github/workflows/tfy-deploy.yaml
```diff
@@ -0,0 +1,26 @@
+name: Deploy
+on:
+  push:
+    branches:
+      - release
+      - main
```
chiragjn, line 6: We need to have different deployments for different branches

truefoundry.yaml
```diff
@@ -30,12 +30,12 @@ components:
...
-        # repo_url: https://github.com/truefoundry/cognita
-        # branch_name: ps_cleanup_metadatastore
-        local_build: false
-        type: local
+        ref: fdfce2e13d5939d7d1c1fce57b90748d8cadd65b
+        type: git
+        repo_url: https://github.com/truefoundry/cognita
+        branch_name: ps_cleanup_metadatastore
+        # local_build: false
+        # type: local
```
chiragjn, line 38: Better to remove commented code

PR #337, opened 2024-09-14
Added Video Parser and Whisper audio parser

S1LV3RJ1NX:


sample-data/video/rag_intro.mp4
chiragjn, line 1: Let's not add these huge files to git tree please, if needed we can put them on our Cloudfront S3 and just add a script to download in this folder

truefoundry.yaml
```diff
@@ -11,7 +11,6 @@ components:
       MODELS_CONFIG_PATH: ./models_config.truefoundry.yaml
       METADATA_STORE_CONFIG: '{"provider":"prisma"}'
       ML_REPO_NAME: cognita-internal
-      INFINITY_URL: http://cas-infinity.cognita-internal.svc.cluster.local:8000
```
chiragjn, line 14: Why is this being removed?

truefoundry.yaml
```diff
@@ -30,12 +29,12 @@ components:
...
-        repo_url: https://github.com/truefoundry/cognita
-        branch_name: ps_cleanup_metadatastore
-        # local_build: false
-        # type: local
+        # ref: fdfce2e13d5939d7d1c1fce57b90748d8cadd65b
+        # type: git
+        # repo_url: https://github.com/truefoundry/cognita
+        # branch_name: ps_cleanup_metadatastore
+        local_build: false
+        type: local
```
chiragjn, line 37: Why are we commenting / uncommenting different configurations?
If needed, let's keep it local build only

Github workflow can be adjusted to replace entire build_source

truefoundry.yaml
```diff
@@ -401,3 +412,51 @@ components:
...
+  - name: cas-whisper
+    env:
+      WHISPER_PORT: 8000
+      WHISPER_MODEL: Systran/faster-distil-whisper-large-v3
+    type: service
+    image:
+      type: image
+      image_uri: fedirz/faster-whisper-server:latest-cpu
+    ports:
+      - host: cas-whisper.truefoundry.com
```
chiragjn, line 426: How was this host decided?
actually we should not even be exposing this, as we don't intend to hit it directly

Please set `expose: false`

docker-compose.yaml
```diff
@@ -129,6 +129,27 @@ services:
...
+    container_name: faster-whisper
+    restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8000/health"]
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
+    environment:
+      - WHISPER__MODEL=${WHISPER_MODEL}
```
chiragjn, line 145: Is the double underscore intentional?

docker-compose.yaml
```diff
@@ -177,7 +198,7 @@ services:
       - BRAVE_API_KEY=${BRAVE_API_KEY}
       - CARBON_AI_API_KEY=${CARBON_AI_API_KEY}
     entrypoint: /bin/bash
-    command: -c "set -e; prisma db push --schema ./backend/database/schema.prisma && uvicorn --host 0.0.0.0 --port 8000 backend.server.app:app --reload"
+    command: -c "set -e; prisma db push --accept-data-loss --schema ./backend/database/schema.prisma && uvicorn --host 0.0.0.0 --port 8000 backend.server.app:app --reload"
```
chiragjn, line 201: Why would we want to accept data loss?
I don't think this should be the case always

chiragjn: In my opinion, we should stop using `db push` and rather use `prisma migrate` and follow proper migrations process

backend/types.py
```diff
@@ -157,7 +159,7 @@ class ParserConfig(ConfiguredBaseModel):
     Parser configuration
     """
 
-    name: str
+    name: str  # File extension
```
chiragjn, line 162: What does this comment mean?

backend/modules/model_gateway/model_gateway.py
```diff
@@ -11,6 +11,7 @@
 from backend.settings import settings
 from backend.types import ModelConfig, ModelProviderConfig, ModelType
 
+from .audio_processing_svc import AudioProcessingSvc
 from .reranker_svc import InfinityRerankerSvc
```
chiragjn, line 15: nit, use absolute imports

backend/modules/model_gateway/model_gateway.py
```diff
@@ -159,5 +178,22 @@ def get_reranker_from_model_config(self, model_name: str, top_k: int = 3):
...
+        if model_name not in self.model_name_to_provider_config:
+            raise ValueError(f"Model {model_name} not registered in the model gateway.")
+        model_provider_config: ModelProviderConfig = self.model_name_to_provider_config[
+            model_name
+        ]
+        if not model_provider_config.api_key_env_var:
+            api_key = "EMPTY"
+        else:
+            api_key = os.environ.get(model_provider_config.api_key_env_var, "")
+        model_id = "/".join(model_name.split("/")[1:])
```
chiragjn, line 191: `_, model = model_name.split("/", 1)`

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def _get_audio_from_video(self, video_filepath: str) -> str:
+        """
+        Extract audio from video
+        """
+        _, file_name = os.path.split(video_filepath)
+        # Save the audio in a temporary directory
+        temp_audio_dir = tempfile.mkdtemp()
+        clip = VideoFileClip(video_filepath)
+        clip.audio.write_audiofile(os.path.join(temp_audio_dir, file_name + ".wav"))
```
chiragjn, line 103: Is there a async interface for this?

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+        return temp_images_dir
+
+    def _get_audio_from_video(self, video_filepath: str) -> str:
+        """
+        Extract audio from video
+        """
+        _, file_name = os.path.split(video_filepath)
+        # Save the audio in a temporary directory
+        temp_audio_dir = tempfile.mkdtemp()
```
chiragjn, line 101: Who will clean this up?

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+            os.path.join(temp_images_dir, file_name + "_frame%04d.png"),
+            fps=0.2,  # configure this for controlling frame rate.
+        )
+        return temp_images_dir
+
+    def _get_audio_from_video(self, video_filepath: str) -> str:
+        """
+        Extract audio from video
+        """
+        _, file_name = os.path.split(video_filepath)
```
chiragjn, line 99: Use os.path.basename

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+
+        super().__init__(**kwargs)
+
+    def _get_images_from_video(self, video_filepath: str) -> str:
+        """
+        Extract images from video frames,
+        We extract 1 frame every 5 seconds of the video
+        Save all the images in temporary directory
+        """
+        _, file_name = os.path.split(video_filepath)
```
chiragjn, line 85: Use os.path.basename

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+        file_name, _ = os.path.splitext(file_name)
+        clip = VideoFileClip(video_filepath)
+        temp_images_dir = tempfile.mkdtemp()
```
chiragjn, line 88: This won't be cleaned up, instead a temp working dir should be passed to this function

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+                max_chunk_size=self.max_chunk_size,
+            )
+
+            # Extract text from audio
+            audio_text = await audio_parser.get_chunks(
+                os.path.join(temp_audio_dir, os.listdir(temp_audio_dir)[0]), metadata
+            )
+
+            # Combine all the extracted text
+            final_text = parsed_video_text + audio_text
```
chiragjn, line 146: Are we preserving metadata correctly? as in what file, modality, timestamp (in case of audio and video) the text came from?

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+            # Remove temporary directories
+            try:
+                for temp_dir in [temp_images_dir, temp_audio_dir]:
+                    for file in os.listdir(temp_dir):
+                        os.remove(os.path.join(temp_dir, file))
+                    os.rmdir(temp_dir)
+                    logger.info(f"Removed temporary directory: {temp_dir}")
+            except Exception as e:
+                logger.exception(f"Error in removing temporary directories: {e}")
+
```
chiragjn, line 157: Instead of manually managing these, use tempfile's context manager, to create a working directory that will be auto cleaned up before exiting get_chunks scope


```
async get_chunks(...):
    with tempfile.TemporaryDirectory(...) as directory:
        try:
            await self._get_chunks(...)
        ...

File: backend/modules/parsers/videoparser.py [hunk 16]
```diff
@@ -0,0 +1,162 @@
...
+        We extract 1 frame every 5 seconds of the video
+        Save all the images in temporary directory
+        """
+        _, file_name = os.path.split(video_filepath)
+        file_name, _ = os.path.splitext(file_name)
+        clip = VideoFileClip(video_filepath)
+        temp_images_dir = tempfile.mkdtemp()
+        clip.write_images_sequence(
+            os.path.join(temp_images_dir, file_name + "_frame%04d.png"),
+            fps=0.2,  # configure this for controlling frame rate.
```
chiragjn, line 91: This should move to config

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Get the chunks of the video file.
+        """
+        try:
+            # Extract images from video
+            temp_images_dir = self._get_images_from_video(filepath)
+
+            # Init multimodal parser for extracting text from images
+            multimodal_parser = MultiModalParser(
+                model_configuration=self.vlm_model_configuration, prompt=self.prompt
```
chiragjn, line 118: max_chunk_size is not being propagated here?

backend/modules/model_gateway/audio_processing_svc.py
```diff
@@ -0,0 +1,60 @@
...
+            total=3,
+            backoff_factor=1,
+            status_forcelist=[429, 500, 502, 503, 504],
+            allowed_methods=["POST"],
+        )
+        self.adapter = HTTPAdapter(max_retries=self.retry_strategy)
+        self.session.mount("https://", self.adapter)
+        self.session.mount("http://", self.adapter)
+
+    def get_transcription(self, audio_file_path: str) -> str:
```
chiragjn, line 42: Return type annotation is incorrect

backend/modules/parsers/audioparser.py
```diff
@@ -0,0 +1,112 @@
...
+            response = self.audio_processing_svc.get_transcription(
+                audio_file_path=filepath
+            )
+
+            try:
+                for line in response.iter_lines():
+                    if line:
+                        data = json.loads(line.decode("utf-8").split("data: ")[1])[
+                            "text"
+                        ]
```
chiragjn, line 76: Maybe we should just add sseclient-py as dependency for all SSE streaming
https://github.com/mpetazzoni/sseclient

truefoundry.yaml
S1LV3RJ1NX: the url is present in the config.yaml file hence.

truefoundry.yaml
S1LV3RJ1NX: Sure, I'll keep it to local, we can modify the github workflow

docker-compose.yaml
S1LV3RJ1NX: this was a typo, will fix.

docker-compose.yaml
S1LV3RJ1NX: this was used when I a another branch had a different schema, we'll remove that for now, will do the migration PR later.

backend/types.py
S1LV3RJ1NX: the name here denotes the file extension when writing the parser config, this was just to identify that.

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: Did not find any.
Ref:
- https://moviepy-tburrows13.readthedocs.io/en/improve-docs/ref/VideoClip/VideoFileClip.html
- https://www.llamaindex.ai/blog/multimodal-rag-for-advanced-video-processing-with-llamaindex-lancedb-33be4804822e

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: yes, the metadata is preserved, for video parser the frame no, audio parser does not return the timestamp. Having that would have been better. 
Ref: https://github.com/fedirz/faster-whisper-server/tree/master

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: Sure, would move to context manager.

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Extract images from video frames,
+        We extract 1 frame every 5 seconds of the video
+        Save all the images in temporary directory
+        """
+        _, file_name = os.path.split(video_filepath)
+        file_name, _ = os.path.splitext(file_name)
+        clip = VideoFileClip(video_filepath)
+        temp_images_dir = tempfile.mkdtemp()
+        clip.write_images_sequence(
```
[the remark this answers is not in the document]
S1LV3RJ1NX: did not find any.

backend/modules/parsers/audioparser.py
S1LV3RJ1NX: will consider this for future.

backend/modules/parsers/audioparser.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        try:
+            parsed_audio_text = []
+
+            response = self.audio_processing_svc.get_transcription(
+                audio_file_path=filepath
+            )
+
+            try:
+                for line in response.iter_lines():
```
chiragjn, line 72: This should ideally be async, done with aiohttp for asyncio

sample-data/video/rag_intro.mp4
S1LV3RJ1NX: Where can I get the access to cloudfront S3 and script to download those files? These sample audio and video files are kept for local users as well to test. Will purge from git tree later.

truefoundry.yaml
S1LV3RJ1NX: Done,
urls are currently determined based on where we are deploying this yaml was for prod cluster. 
We are in process of generating a `deploy.py` file, which should take `workspace_fqn` and cluster `base_domain_url` as args, this should solve the issue soon.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: done.

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: done.

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: done.

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: done.

backend/modules/parsers/videoparser.py
S1LV3RJ1NX: Currently we are not doing any chunking in `MultiModalParser`. 
We can later on gather all text and pass it to `UnstructuredIoParser`

backend/modules/parsers/audioparser.py
```diff
@@ -0,0 +1,112 @@
...
+                tempfile_name = temp_file.name
+
+            # Split the text into chunks
+            unstructured_io_parser = UnstructuredIoParser(
+                max_chunk_size=self.max_chunk_size
+            )
+
+            final_texts = await unstructured_io_parser.get_chunks(
+                filepath=tempfile_name, metadata=metadata
+            )
```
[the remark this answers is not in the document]
S1LV3RJ1NX: we are not yet on `3.12` we use `3.11`

backend/modules/parsers/audioparser.py
S1LV3RJ1NX: done.

sample-data/video/rag_intro.mp4
chiragjn: We have assets.truefoundry.com, we can create a folder for cognita in that same bucket allow downloading using something like `https://assets.truefoundry.com/cognita/sample_data/...`

Just need to add a simple README / bash script that can wget this url

chiragjn: Let me upload and get back here

truefoundry.yaml
```diff
@@ -401,3 +395,51 @@ components:
         persistence:
           size: 5Gi
       architecture: standalone
+
+  # Whisper
+  - name: cas-whisper
+    env:
+      WHISPER_PORT: 8000
+      WHISPER__MODEL: Systran/faster-distil-whisper-large-v3
+      WHISPER__INFERENCE_DEVICE: auto
```
chiragjn, line 404: These double underscores in key names are correct?

truefoundry.yaml
chiragjn: Please add a ticket for this

S1LV3RJ1NX: Done.

truefoundry.yaml
S1LV3RJ1NX: yes, they are from the dockerfile, checked.

backend/modules/model_gateway/model_gateway.py
```diff
@@ -159,5 +177,22 @@ def get_reranker_from_model_config(self, model_name: str, top_k: int = 3):
             top_k=top_k,
         )
 
+    async def get_audio_model_from_model_config(self, model_name: str):
```
chiragjn, line 180: This does not need to be async def, right?
I don't see any async call in this method itself

truefoundry.yaml
S1LV3RJ1NX: https://github.com/fedirz/faster-whisper-server/blob/master/Dockerfile.cpu

backend/modules/model_gateway/model_gateway.py
S1LV3RJ1NX: It is used in `AudioParser` service.

backend/modules/model_gateway/audio_processing_svc.py
```diff
@@ -0,0 +1,69 @@
...
+            "temperature": 0.1,
+            "response_format": "json",
+            "language": "en",
+            "timestamp_granularities": "segment",
+            "stream": "true",
+        }
+
+    # allows class to be used as async context manager
+    async def __aenter__(self):
+        self.session = aiohttp.ClientSession()
```
chiragjn, line 33: We need to add a todo to see how to do retries with this

backend/modules/model_gateway/audio_processing_svc.py
```diff
@@ -0,0 +1,69 @@
...
+            await self.session.close()
+
+    async def get_transcription(self, audio_file_path: str) -> aiohttp.ClientResponse:
+        """
+        Get streaming audio transcription from Faster-Whisper Server
+        """
+        if not self.session:
+            raise RuntimeError(
+                "Session not initialized. Use 'async with' context manager."
+            )
```
chiragjn, line 48: Is there really a use case to make this class itself async context manager?
Would like to keep things simpler if we can

backend/modules/model_gateway/model_gateway.py
chiragjn: Is this because AudioProcessingSvc has been made into a async context manager?

backend/modules/parsers/audioparser.py
```diff
@@ -0,0 +1,125 @@
...
+
+    async def initialize(self):
+        """
+        Initialize the audio processing service.
+        """
+        self.audio_processing_svc = (
+            await model_gateway.get_audio_model_from_model_config(
+                model_name=self.model_configuration.name
+            )
+        )
```
chiragjn, line 64: This does not really have to be async if we keep get_audio_model_from_model_config sync, then can be part of __init__ itself

S1LV3RJ1NX: yaa, now that it is async, I cant use it in init. Will refactor this later.

backend/modules/parsers/videoparser.py
```diff
@@ -0,0 +1,162 @@
...
+    async def get_chunks(
+        self, filepath: str, metadata: Dict[Any, Any] | None, **kwargs
+    ) -> List[Document]:
+        """
+        Get the chunks of the video file.
+        """
+        with tempfile.TemporaryDirectory(suffix="") as directory:
+            try:
+                # Extract images from video
+                temp_images_dir = os.path.join(directory, "images")
```
chiragjn, line 115: nitpick, we can refactor main logic to another function to reduce indent

backend/modules/model_gateway/audio_processing_svc.py
```diff
@@ -0,0 +1,69 @@
...
+    async def __aenter__(self):
+        self.session = aiohttp.ClientSession()
+        return self
+
+    # allows class to be used as async context manager
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        if self.session:
+            await self.session.close()
+
+    async def get_transcription(self, audio_file_path: str) -> aiohttp.ClientResponse:
```
chiragjn, line 41: Why not make this `AsyncIterator[str]` ?
iterate the response and yield lines here itself

That way you can also close the session on once all lines are done, no need for class to be async context manager

backend/modules/parsers/audioparser.py
```diff
@@ -0,0 +1,125 @@
...
+                    audio_file_path=filepath
+                )
+
+                try:
+                    async for line in response.content:
+                        line = line.strip()
+                        if line:
+                            data = json.loads(
+                                line.decode("utf-8").strip().split("data: ")[1]
+                            )["text"]
```
chiragjn, line 89: see comment about moving this to audio service itself
https://github.com/truefoundry/cognita/pull/337/files#r1765351573

chiragjn: this can just become

```
async for text_chunk in self.audio_processing_svc.get_transcription(...):
    parsed_audio_text.append(text_chunk)
```

chiragjn (approved):
Approving to unblock,
Main thing to address: 

https://github.com/truefoundry/cognita/pull/337/files#r1765351573
https://github.com/truefoundry/cognita/pull/337/files#r1765356099

backend/modules/parsers/audioparser.py
```diff
@@ -46,52 +46,35 @@ def __init__(
...
-                        line = line.strip()
-                        if line:
-                            data = json.loads(
-                                line.decode("utf-8").strip().split("data: ")[1]
-                            )["text"]
-                            parsed_audio_text.append(data)
-                            logger.info(f"Transcription: {data}")
+                    data = json.loads(line)["text"]
+                    parsed_audio_text.append(data)
+                    logger.info(f"Transcription: {data}")
```
chiragjn, line 71: This being in the loop might be too spamy, dedent one level

backend/modules/parsers/audioparser.py
```diff
@@ -46,52 +46,35 @@ def __init__(
...
+                except json.JSONDecodeError:
+                    logger.error(f"Error decoding JSON: {line}")
+                except KeyError:
+                    logger.error(f"Missing 'text' key in JSON: {line}")
```
chiragjn, line 75: Not sure if we should be catching and skipping chunks like this

backend/modules/parsers/audioparser.py
```diff
@@ -0,0 +1,111 @@
...
+            async for line in self.audio_processing_svc.get_transcription(filepath):
+                try:
+                    data = json.loads(line)["text"]
+                    parsed_audio_text.append(data)
+                except json.JSONDecodeError:
+                    logger.exception(f"Error decoding JSON: {line}")
+                    raise json.JSONDecodeError
+                except KeyError:
+                    logger.exception(f"Missing 'text' key in JSON: {line}")
+                    raise KeyError
```
chiragjn:
```suggestion
                except json.JSONDecodeError as je:
                    logger.exception(f"Error decoding JSON: {line}")
                    raise je
                except KeyError as ke:
                    logger.exception(f"Missing 'text' key in JSON: {line}")
                    raise ke
```

chiragjn (approved):

Issue #330, opened 2024-09-08
Indexer Issue | FastApi Process Pool is re-used ? 


jalotra:
hi dev, I am not able to understand the indexing flow, 
these are logs : 
```
cognita-backend   | DEBUG:    2024-09-08 12:43:44,504 - indexer:ingest_data:342 - Starting ingestion for data source fqn: localdir::/app/user_data/law-pdf
cognita-backend   | INFO:     192.168.65.1:33019 - "POST /v1/collections/ingest HTTP/1.1" 201 Created
cognita-backend   | INFO:     192.168.65.1:33019 - "GET /v1/collections/law-pdf-machine HTTP/1.1" 200 OK
cognita-backend   | INFO:     192.168.65.1:33019 - "POST /v1/collections/data_ingestion_runs/list HTTP/1.1" 200 OK
cognita-backend   | ERROR:    2024-09-08 12:43:44,617 - prismastore:aupdate_data_ingestion_run_status:551 - Failed to update data ingestion run status: Event loop is closed
cognita-backend   | Traceback (most recent call last):
cognita-backend   |   File "/app/backend/modules/metadata_store/prismastore.py", line 538, in aupdate_data_ingestion_run_status
cognita-backend   |     ] = await self.db.ingestionruns.update(
```

when I look at what `/v1/collections/ingest` does is this : 
```
try:
        process_pool = request.app.state.process_pool
    except AttributeError:
        process_pool = None
  ```
  
  The issue I think looks like this : 
1. If anything breaks in previous runs, the event loop `breaks out` and then successive runs are not possible

stack : 
```
cognita-backend   |   File "/virtualenvs/venv/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 216, in handle_async_request
cognita-backend   |     raise exc from None
cognita-backend   |   File "/virtualenvs/venv/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 189, in handle_async_request
cognita-backend   |     await self._close_connections(closing)
cognita-backend   |   File "/virtualenvs/venv/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 305, in _close_connections
cognita-backend   |     await connection.aclose()
cognita-backend   |   File "/virtualenvs/venv/lib/python3.11/site-packages/httpcore/_async/connection.py", line 171, in aclose
cognita-backend   |     await self._connection.aclose()
[...]

PR #342, opened 2024-09-18
# Improved Process Pool Management and Error Handling

Malyala-Srikanth:
## Changes
- Refactored `AsyncProcessPoolExecutor` to correctly handle coroutines and awaitables
- Implemented an asynchronous `submit` method in `AsyncProcessPoolExecutor`
- Added a periodic health check mechanism in the process pool lifespan manager
- Introduced an automatic restart feature for the process pool on detected failures
- Enhanced exception handling and logging for tasks submitted to the process pool

## Motivation
These changes address stability issues we've been experiencing with process pool, particularly when tasks raise unhandled exceptions. The new implementation provides better visibility into errors occurring within pool tasks and implements a self-healing mechanism to maintain system stability.

## Key Improvements
1. **Proper Coroutine Handling**: The `AsyncProcessPoolExecutor` now correctly manages coroutines, aligning with our async-first architecture.
2. **Health Monitoring**: Regular health checks ensure early detection of pool failures.
3. **Auto-Recovery**: The pool now automatically restarts if it enters a broken state, reducing downtime and manual intervention.
4. **Enhanced Logging**: Improved error logging provides better insights for debugging and monitoring.

#Testing.

Hi Team, I was encountering the same issue as in #330 and able to reproduce by sending an ingestion request with incorrect params
```
"model_configuration": {
        "name": "random"
      }
```  
without the model registered which was leading to an error and responsible for breaking the pool. 

Please review and provide feedback, particularly on the error handling strategy and the frequency of health checks. Suggest me if there are any changes required.

### DIFF
--- a/backend/server/app.py
+++ b/backend/server/app.py
@@ -1,3 +1,4 @@
+import asyncio
 import multiprocessing as mp
 from contextlib import asynccontextmanager
 
@@ -5,6 +6,7 @@ from fastapi import APIRouter, FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 
+from backend.logger import logger
 from backend.modules.query_controllers.query_controller import QUERY_CONTROLLER_REGISTRY
 from backend.server.routers.collection import router as collection_router
 from backend.server.routers.components import router as components_router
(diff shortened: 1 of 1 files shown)


backend/server/app.py
```diff
@@ -12,15 +13,40 @@
...
         max_workers=settings.PROCESS_POOL_WORKERS
     )
-    yield  # FastAPI runs here
-    app.state.process_pool.shutdown(wait=True)
+
+    async def check_pool_health():
+        while True:
+            try:
+                # Submit a simple task to check if the pool is responsive
+                app.state.process_pool.submit(asyncio.sleep, 0)
```
chiragjn, line 29: We should probably call`.result()` on this future

backend/utils.py
```diff
@@ -91,7 +92,21 @@ def wrapper() -> T:
...
+        return future
+
+    async def _check_future(self, future):
+        try:
+            result = await asyncio.wrap_future(future)
+            if isinstance(result, Exception):
+                logger.error(f"Task in process pool failed: {result}")
+                # You could implement additional error handling here
+        except Exception as e:
+            logger.error(f"Error checking future in process pool: {e}")
```
chiragjn, line 112: I would like to simplify a few things here
---
I was trying to get to the reason why the event loop closes in the first place, because I cannot reproduce it minimally

```
import time
import asyncio
from concurrent.futures import Executor, ProcessPoolExecutor

class AsyncProcessPoolExecutor(ProcessPoolExecutor):
    @staticmethod
    def _async_to_sync(__fn, *args, **kwargs):
        return asyncio.run(__fn(*args, **kwargs))

    def submit(self, __fn, *args, **kwargs):
        return super().submit(self._async_to_sync, __fn, *args, **kwargs)


pool = AsyncProcessPoolExecutor(max_workers=1)

async def crash():
    raise Exception("Crashed!")


async def main():
    futures = []
    for i in range(10):
        print("Submitting ... ")
        futures.append(pool.submit(crash))
        time.sleep(1)
    for fut in futures:
        print(f"{fut.exception()!r}")

if __name__ == "__main__":
    asyncio.run(main())
```

Let me quickly reproduce this with the whole setup

chiragjn:
Thanks for taking the time to contribute!
I have started #349, so we should not need any changes to `AsyncProcessPoolExecutor`

But I still like the robustness changes you have made in FastAPI lifespan, we can definitely get them in. Let's rebase once we have #349 merged

chiragjn (approved):

chiragjn:
Changes seem good to go, can you please run pre-commit once?

PR #350, opened 2024-09-20
Upgrade react and daisyui

Blakeinstein:
- **Implement carbon (squashed commit)**
- **revert carbon**


backend/indexer/indexer.py
```diff
@@ -259,29 +259,6 @@ async def ingest_data_points(
             documents_to_be_upserted.append(chunk)
         logger.info("%s -> %s chunks", loaded_data_point.local_filepath, len(chunks))
 
-        # delete the file from temp dir after processing
```
S1LV3RJ1NX, line 262: Why was this removed?

backend/indexer/indexer.py
```diff
@@ -367,15 +344,9 @@ async def ingest_data(
                     batch_size=request.batch_size,
                 )
                 if pool:
-                    logger.info(
```
S1LV3RJ1NX, line 370: Good to have these logging statements.

S1LV3RJ1NX (commented):
minor doubts.

backend/indexer/indexer.py
```diff
@@ -259,29 +259,6 @@ async def ingest_data_points(
...
-            if loaded_data_point.local_metadata_file_path:
-                os.remove(loaded_data_point.local_metadata_file_path)
-                logger.debug(
-                    f"Processing done! Deleting file {loaded_data_point.local_metadata_file_path}"
-                )
-        except Exception as e:
-            logger.exception(
-                f"Failed to delete file {loaded_data_point.local_metadata_file_path} after processing. Error: {e}"
-            )
-
```
S1LV3RJ1NX, line 284: Why was this removed?

S1LV3RJ1NX (commented):
updated the comment

backend/indexer/indexer.py
Blakeinstein: The temp directory from the library is auto-cleaned when the variable is out of scope. So, this code was redundant.

sayan-truefoundry:
did we change the style of this input?
image

sayan-truefoundry:
also i would suggest pls give the font color white to "Cancel"
image

sayan-truefoundry:
create new data source flow has bugs now ... 
i can see after creating a valid data source using web, the drawer didn't get closed and after that i have hit the submit button again and a blank data source got created

sayan-truefoundry:
in the embedded application the submit button looks faded now ... pls bring back the previous black color
image

sayan-truefoundry:
image
 i will suggest make the text color white whenever this Cancel button is in hover state and it would be great if we could have the black background in Create button by default rather than this violet or indigio (we use in all other places)

sayan-truefoundry:
in data source creation form, this now works fine and closes the drawer after successful data source creation that is fine.... but after that if i open the drawer again without refreshing then no tab is shown as selected in Data Source Type 
image

sayan-truefoundry (approved):

Issue #363, opened 2024-10-09
Parser Registry Problem


magaton:
Hello, I am looking at the code and it seems that a new Parser impl instance is created for each file in the datasource.
For 100K files repos this looks like a significant overhead. Disclaimer: I am not a Python dev, just a random observer :)

https://github.com/truefoundry/cognita/blob/01b22433b6a25973d8bf561a23c619c5f41770e7/backend/indexer/indexer.py#L234

and 

https://github.com/truefoundry/cognita/blob/main/backend/modules/parsers/parser.py#L91

Shouldn't the parser instances be created per datasource and not cached globally?
WDYT?

chiragjn:
Thanks for reporting this, you are correct
We should actually be caching instances for the entire duration of indexing

PR #374, opened 2024-10-12
feat: add read signed url of the vector chunks

mnvsk97:
- Add a function to fetch artifacts of the data directory by data dir fqn.
- Add function to get `read signed URL` of a data dir and local file path
- Add `_signed_url` in the Q/A API response to enable UI to reference and provide links to the source files.
- Handle metadata enrichment for stream and nonstream response types.
- Refactor `MultiModalParser`
  - Move default multi-modal parser prompt to constants
  - Move LLM instance creation to `__init__`
  - Remove unnecessary passing of instance objects like prompt, and llm between functions and use instance vars instead.
- Upgrade `langchain-openai` from `0.1.20` to `0.1.25` to fix issues introduced in https://github.com/truefoundry/cognita/pull/371
- Remove unused libraries: `openai` and `orjson`
- Closes https://github.com/truefoundry/cognita/issues/363

### DIFF
--- a/backend/constants.py
+++ b/backend/constants.py
@@ -2,8 +2,31 @@ DATA_POINT_FQN_METADATA_KEY = "_data_point_fqn"
 
 DATA_POINT_HASH_METADATA_KEY = "_data_point_hash"
 
+DATA_POINT_SIGNED_URL_METADATA_KEY = "_signed_url"
+
+DATA_POINT_FILE_PATH_METADATA_KEY = "_data_point_file_path"
+
 DEFAULT_BATCH_SIZE = 100
 
 DEFAULT_BATCH_SIZE_FOR_VECTOR_STORE = 1000
 
 FQN_SEPARATOR = "::"
+
+# parser constants
+
+MULTI_MODAL_PARSER_PROMPT = """Given an image containing one or more charts/graphs, and texts, provide a detailed analysis of the data represented in the charts. Your task is to analyze the image and provide insights based on the data it represents.
+    Specifically, the information should include but not limited to:
+    Title of the Image: Provide a title from the charts or image if any.
+    Type of Chart: Determine the type of each chart (e.g., bar chart, line chart, pie chart, scatter plot, etc.) and its key features (e.g., labels, legends, data points).
+    Data Trends: Describe any notable trends or patterns visible in the data. This may include increasing/decreasing trends, seasonality, outliers, etc.
+    Key Insights: Extract key insights or observations from the charts. What do the charts reveal about the underlying data? Are there any significant findings that stand out?
+    Data Points: Identify specific data points or values represented in the charts, especially those that contribute to the overall analysis or insights.
+    Comparisons: Compare different charts within the same image or compare data points within a single chart. Highlight similarities, differences, or correlations between datasets.
+    Conclude with a summary of the key findings from your analysis and any recommendations based on those findings.
+"""
+MULTI_MODAL_PARSER_SUPPORTED_IMAGE_EXTENSIONS = [".png", ".jpeg", ".jpg"]
+MULTI_MODAL_PARSER_SUPPORTED_PDF_EXTENSION = [".pdf"]
+MULTI_MODAL_PARSER_SUPPORTED_FILE_EXTENSIONS = (
+    MULTI_MODAL_PARSER_SUPPORTED_IMAGE_EXTENSIONS
+    + MULTI_MODAL_PARSER_SUPPORTED_PDF_EXTENSION
+)
--- a/backend/modules/parsers/multi_modal_parser.py
+++ b/backend/modules/parsers/multi_modal_parser.py
@@ -11,6 +11,12 @@ from langchain.docstore.document import Document
 from langchain_core.language_models.chat_models import BaseChatModel
 from langchain_core.messages import HumanMessage
 
+from backend.constants import (
+    MULTI_MODAL_PARSER_PROMPT,
+    MULTI_MODAL_PARSER_SUPPORTED_FILE_EXTENSIONS,
+    MULTI_MODAL_PARSER_SUPPORTED_IMAGE_EXTENSIONS,
+    MULTI_MODAL_PARSER_SUPPORTED_PDF_EXTENSION,
+)
 from backend.logger import logger
 from backend.modules.model_gateway.model_gateway import model_gateway
 from backend.modules.parsers.parser import BaseParser
@@ -20,7 +26,7 @@ from backend.types import ModelConfig
 
 class MultiModalParser(BaseParser):
     """
-    MultiModalParser is a multi-modal parser class for deep extraction of pdf documents.
+    MultiModalParser is a multi-modal parser class for deep extraction of pdf documents and images.
 
     Parser Configuration will look like the following while creating the collection:
     {
@@ -36,7 +42,7 @@ class MultiModalParser(BaseParser):
     }
     """
 
-    supported_file_extensions = [".pdf", ".png", ".jpeg", ".jpg"]
+    supported_file_extensions = MULTI_MODAL_PARSER_SUPPORTED_FILE_EXTENSIONS
 
     def __init__(self, *, model_configuration: ModelConfig, prompt: str = "", **kwargs):
         """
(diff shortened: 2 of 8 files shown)


backend/constants.py
```diff
@@ -2,8 +2,29 @@
...
+    Specifically, the information should include but not limited to:
+    Title of the Image: Provide a title from the charts or image if any.
+    Type of Chart: Determine the type of each chart (e.g., bar chart, line chart, pie chart, scatter plot, etc.) and its key features (e.g., labels, legends, data points).
+    Data Trends: Describe any notable trends or patterns visible in the data. This may include increasing/decreasing trends, seasonality, outliers, etc.
+    Key Insights: Extract key insights or observations from the charts. What do the charts reveal about the underlying data? Are there any significant findings that stand out?
+    Data Points: Identify specific data points or values represented in the charts, especially those that contribute to the overall analysis or insights.
+    Comparisons: Compare different charts within the same image or compare data points within a single chart. Highlight similarities, differences, or correlations between datasets.
+    Conclude with a summary of the key findings from your analysis and any recommendations based on those findings.
+"""
+MULTI_MODAL_PARSER_SUPPORTED_IMAGE_EXTENSIONS = [".png", ".jpeg", ".jpg"]
```
S1LV3RJ1NX, line 25: why was this necessary to to move it out as constant? It's not being used anywhere else right?

backend/modules/parsers/multi_modal_parser.py
```diff
@@ -12,6 +12,12 @@
 from langchain_core.messages import HumanMessage
 from PIL import Image
 
+from backend.constants import (
+    MULTI_MODAL_PARSER_PROMPT,
+    MULTI_MODAL_PARSER_SUPPORTED_FILE_EXTENSIONS,
+    MULTI_MODAL_PARSER_SUPPORTED_IMAGE_EXTENSIONS,
+    MULTI_MODAL_PARSER_SUPPORTED_PDF_EXTENSION,
+)
```
S1LV3RJ1NX, line 20: Not sure why this was done only for multimodal parser, specifically constants for managing extensions.

backend/modules/parsers/multi_modal_parser.py
```diff
@@ -87,129 +82,89 @@ async def call_vlm_agent(
...
+        self, file_name: str, pg_no: int, page_content: str, image_b64: str
+    ) -> Document:
+        return Document(
+            page_content=f"File Name: {file_name}\n\n{page_content}",
+            metadata={
+                "image_b64": image_b64,
+                "page_number": pg_no,
+                "source": file_name,
+            },
+        )
```
S1LV3RJ1NX, line 170: I see that pre-signed uri is added post parsing stage. when chunks are returned. Just a thought, would it be beneficial to add the signed uri here it self, even if we don't, there is one missing step, we are not updating the chunk metadata with document metadata `_metadata` arg from `get_chunks` function. 

Please check for other parsers as well.

backend/modules/parsers/parser.py
```diff
@@ -54,6 +54,29 @@ async def get_chunks(
...
+        **kwargs: Arbitrary keyword arguments.
+
+    Returns:
+        BaseParser: The parser for the given extension, either from cache or newly created.
+    """
+    if extension not in cache:
+        cache[extension] = get_parser_for_extension(
+            extension, parsers_map, *args, **kwargs
+        )
+    return cache[extension]
```
S1LV3RJ1NX, line 77: one thing that is not clear to me is, how will two parsers with same extension but different config params will be cached ? Usecase being when the collection has multiple datasources, each data sources uses the same parser but with different config.

S1LV3RJ1NX (changes requested):
Need clarity on two things:
- Need to update chunk metadata with document metadata, if any for all the parsers
- How does caching work for parsers with same extension but different config, as the caching key is the file extension.

mnvsk97:
> Need clarity on two things:
> 
> - Need to update chunk metadata with document metadata, if any for all the parsers
> 
> - How does caching work for parsers with same extension but different config, as the caching key is the file extension.

1. Not sure about this, need an example. 
2. I missed the part around config. I'll add an md5 hash of the config and extension.

S1LV3RJ1NX:
> Not sure about this, need an example.
@mnvsk97 - if you would check get chunks function of any parser, it takes `file_path` and `metadata` lets call this as `doc_metadata` as input argument. So when an individual document chunk generates it's own metadata, we should also add the `doc_metadata` to the chunk metadata.

Another thing that is missing is, 
QueryController's `required_metadata` list should be updated to send pre-signed urls in the response.
https://github.com/mnvsk97/cognita/blob/16c20fb006e9065cc435439d77a44bd58e3981a6/backend/modules/query_controllers/base.py#L24

backend/modules/parsers/multi_modal_parser.py
mnvsk97: Since this PR had changes related to this parser, I moved those constants to a constants.py file.

backend/constants.py
mnvsk97: It's a common pattern to store constants in a single place and import them wherever necessary. In the future, we can rename and re-use such constants as and when needed. Even if not, it's a good organisational pattern. 

Ex: https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/redis/constants.py

For now, we have a common file for constants at backend/constants.py. In the future, we can have a constant file per module.

backend/modules/parsers/multi_modal_parser.py
Blakeinstein: When I implemented this step, I was getting a signed uri on demand using the internal fqn for the data point, the frontend preview code was designed that way too was wondering if we should change that. Moving forward additional data sources may not live on our own storage providers either right? for example external airtable or SQL dbs etc.

mnvsk97: Generating a pre-signed read URL in case of a tfy data dir required tfy fqn, which is not available in this function and will require many changes. Also, this will involve changing all the parsers. I don't think that's a good idea.

backend/modules/parsers/parser.py
mnvsk97: You are right. I've changed the cache key in the latest commits to include parser_config.

backend/modules/parsers/multi_modal_parser.py
S1LV3RJ1NX: @mnvsk97 - I suppose the chunk metadata already has an fqn, you might want to checked the indexed chunks within the qdrant that should help generate pre-signed urls in the query controller step itself. 

As now I see, there is a big flaw here being, the pre-signed url generated in indexing stage becomes part of chunk metadata embedded in the qdrant, as a result signed-url does not change. Hence, after sometime, the pre-signed url will become invalid.

A good way is to - 
- Remove signed url generation code from the indexing and parsing stage
- Add a `_generate_pre_signed_urls` function to `BaseQueryController`, this function is called at the last step of querycontroller when `answer` and `docs` are generated, that uses doc_fqn to generate the url
- Send this pre-signed url as `url` key in the metadata for each chunk when returning the answers
- If you see fqn, is not present, we can add the code for that in data-loading stage and update the chunk metadata with document metadata in the parsing stage.

mnvsk97: Nice catch. I'll move this to `QueryController`

mnvsk97:
> > Not sure about this, need an example.
> 
> @mnvsk97 - if you would check `get_chunks` function of any parser, it takes `file_path` and `metadata` (lets call this as `doc_metadata` for sake of explanation) as input argument. So when an individual document chunk generates it's own metadata, we should also add the `doc_metadata` to the chunk metadata.
> 
> Another thing that is missing is, QueryController's `required_metadata` list should be updated to send pre-signed urls in the response. https://github.com/mnvsk97/cognita/blob/16c20fb006e9065cc435439d77a44bd58e3981a6/backend/modules/query_controllers/base.py#L24

Resolved all the above mentioned points in the latest commits

backend/modules/parsers/multi_modal_parser.py
mnvsk97: Resolved in latest commits

S1LV3RJ1NX (approved):

PR #387, opened 2024-10-22
feat: validate data dir before creating data source

mnvsk97:


backend/server/routers/data_source.py
```diff
@@ -37,9 +37,21 @@ async def list_data_sources():
 async def add_data_source(data_source: CreateDataSource):
     """Create a data source for the given collection"""
     metadata_store_client: BaseMetadataStore = await get_client()
+    # Validate URI before creating the data source
+    if data_source.type == "truefoundry":
```
Blakeinstein, line 41: I thought we had source type constants? would prefer to avoid arbitrary string comparison

mnvsk97: its not there at the moment

Blakeinstein (approved):

PR #395, opened 2024-10-23
Updated Error Handling

ChoudharyHarish:
Fixed Error fixing by building custom error handler.

@Blakeinstein  and @sayan-truefoundry Can you review the updates.

### DIFF
--- a/frontend/src/utils/error.ts
+++ b/frontend/src/utils/error.ts
@@ -1,0 +1,12 @@
+import notify from "@/components/base/molecules/Notify";
+
+
+export function notifyError(baseMessage: string, err?: any, defaultMessage? : string) {
+  const message =
+  err?.message ||
+  err?.data?.error ||
+    err?.error ||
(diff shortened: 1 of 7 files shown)


frontend/src/utils/error.ts
```diff
@@ -0,0 +1,12 @@
+import notify from "@/components/base/molecules/Notify";
+
+
+export function notifyError(baseMessage: string, err?: any, defaultMessage? : string) {
+  const message =
+  err?.message ||
+  err?.data?.error ||
```
mnvsk97, line 7: as per our last discussion, is this still required?

ChoudharyHarish: Yes rishi told to keep like this so because we didn't actually find the reason why those 2 lines were used for handling but we are using function now so if any future we want to remove we can just remove from here only.

Blakeinstein (approved):

PR #408, opened 2024-11-11
Update unstructured_io.py for empty text

S1LV3RJ1NX:


backend/modules/parsers/unstructured_io.py
```diff
@@ -87,6 +87,8 @@ async def get_chunks(self, filepath: str, metadata: dict, **kwargs):
             parsed_data = response.json()
             for payload in parsed_data:
                 text = payload["text"]
+                if text == None or len(text) == 0:
```
mnvsk97, line 90:
```suggestion
                if not text:
```

This covers both the conditions you added.

mnvsk97 (changes requested):

backend/modules/parsers/unstructured_io.py
S1LV3RJ1NX: Cool, I am just vary of how negation would turn out in cases. But can do.

mnvsk97 (approved):

PR #413, opened 2024-12-04
[Vector Store] Mongo DB Integration

mnvsk97:


chiragjn (approved):

jitenderfoundry (approved):

backend/modules/vector_db/__init__.py
```diff
@@ -7,6 +9,7 @@
 
 SUPPORTED_VECTOR_DBS = {
     "qdrant": QdrantVectorDB,
+    # "mongo": MongoVectorDB,
```
chiragjn, line 12: Is this intentional?

mnvsk97: Yes, Qdrant will always be used by default since its not dependent on creating an external account.

PR #432, opened 2025-08-26
feat: Add ChromaDB vector database support

SharanRP:
- Implement ChromaVectorDB class following BaseVectorDB interface
- Add chromadb and langchain-chroma dependencies
- Support both local and remote ChromaDB configurations
- Add configuration examples in compose.env
- Register ChromaDB in vector database factory


mnvsk97:
@SharanRP Thanks for raising the PR. This is very helpful.
Before submitting the review, can you please add documentation to run Chroma locally? 
Ideally, it should be a part of the docker-compose.yaml file so that the users can choose to run it as and when needed.

LMK in case of questions.

